PR #10, opened 2021-09-06
Elral/issue2

elral:
w/o RotaryEncoder and TicksPerSecond libs.
Please double check FastAcceleration for different encoder types!


MobiFlight-Admin (commented):
This looks like a big change. I am OK if tested thoroughly and if backwards compatible.

Observations:

- What's the benefits of having the source code from the other library now as part of our code?
- I am realizing it is BSD license :(
- 

Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:

- Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
- Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
- Neither the name of the copyright owners nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.

danecreekphotography:
> * What's the benefits of having the source code from the other library now as part of our code?

From over in the MobiFlight issue that's tracking the PlatformIO change @elral mentioned that it saves quite a bit of flash:

> I moved the required functions from button and rotaryencoder libs to MFEncoder.cpp. Therefore these libs are no longer required and I do not expect changes in these libs anymore (at least which would be required for MF). Interesting side effect is that this has reduced the RAM usage by ~210byte and Flash by ~2,1kB!??

elral:
The original encoder lib uses the button lib, which is also RAM consuming and not required.

The original license is a topic, I thought it would be enough to add this in the top of both files (.cpp and .h).
If not I could think about alternatives.

danecreekphotography:
@MobiFlight-Admin My read of the BSD license matches @elral: with the proper attribution/copyright it is ok.

Do you have any other questions/concerns with this change?

PR #11, opened 2021-09-06
Reduce RAM usage

elral:
All Buttons are read in a for loop. Every button stores it's own time when it is read the last time, but for all buttons this is the same time (for loop). To save some RAM usage (60bytes for the proMicro) the "debouncing" can be done in the central function (readButtons() ).
Same could be applied to the encoder function, but is not part of the existing pull request. It will "only" save 12Bytes for the proMicro (4 Encoder * 4Byte(_lastmillis in MFENcoder.h) - 4Byte(mobiflight.cpp))...

### DIFF
--- a/include/MFButton.h
+++ b/include/MFButton.h
@@ -43,7 +43,6 @@ public:
     
 private:
     bool          _state;
-    uint32_t      _last;
     buttonEvent   _handlerList[2];    
 };
 #endif 
--- a/src/MF_Modules/MFButton.cpp
+++ b/src/MF_Modules/MFButton.cpp
@@ -9,17 +9,13 @@ MFButton::MFButton(uint8_t pin, const char * name)
   _pin  = pin;
   _name = name;
   _state = 1;
-  _last = millis();
   pinMode(_pin, INPUT);     // set pin to input
   digitalWrite(_pin, HIGH); // turn on pullup resistors
 }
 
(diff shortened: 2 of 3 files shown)


danecreekphotography (approved):
This is such a nice little change. On the Pro Micro space is at such a premium every little bit helps. I have a separate issue opened to cover the same change for encoder buttons, and will apply a similar change to my work for input shift registers. If nothing else it'll ensure that the code for debounce is consistent across all buttons, which helps with code clarity.

elral:
I just saw that there is a typo in there.
In the mobiflight.cpp#1072 is
`uint32_t _lastButtonRead;`
But it should be:
`uint32_t lastButtonRead;`
(w/o the underline). Applies also to line 1076 and 1077.

danecreekphotography:
If you make the change in your branch and push it the pull request to update automatically.

elral:
Done.
Much to learn with these github things....

MobiFlight-Admin (changes requested):
What a pity that we can't encapsulate the debounce behavior in the button class. Or can we do this elegantly by using a static class variable? We could then set it from the outside once.

If not:
Let's rename lastButtonUpdate and convert the 10 into something like "MF_BUTTON_DEBOUNCE_MS" then it is clearer what this does. Everything else is ok.

PR #35, opened 2021-09-20
Fix embedding the version number at build time

### LINKED ISSUE #34: Version number isn't inserted into the build correctly
The version is getting set to "BUILD_VERSION" instead of the value.

danecreekphotography:
Fixes #34


danecreekphotography:
What an adventure 😂

PR #40, opened 2021-09-21
Enable overriding components via platformio.ini

### LINKED ISSUE #39: Enable overriding components via platformio.ini
The #defines that determine what components are supported by each firmware are currently set in the board header files with
no way to modify them from the command line of the build. This makes it impossible to generate board variants without
modifying the header file.

Fix this by wrapping each #define with an #ifndef, so it provides the default but can be overridden at compile time.

neilenns:
Fixes #39

Wrap the defines that determine what components are supported by the firmware in #ifndef statements. This enables setting them via the `platformio.ini` file in the future to simplify creating firmware variants for a board at build time.


danecreekphotography:
As an example this change will allow something like this to get added to the platformio.ini file:

```
[env:promicro_basic]
platform = atmelavr
board = sparkfun_promicro16
framework = arduino
build_flags = 
	-I./_Boards/Atmel/Board_ProMicro
	-D MF_SEGMENT_SUPPORT=0
	-D MF_LCD_SUPPORT=0
src_filter = 
	+<*>
	+<../_Boards/Atmel>
lib_deps = 
	arduino-libraries/Servo @ ^1.1.7
	wayoda/LedControl @ ^1.0.6
	waspinator/AccelStepper @ ^1.61
	marcoschwartz/LiquidCrystal_I2C @ ^1.1.4
	thijse/EEPROMEx@0.0.0-alpha+sha.09d7586108
monitor_speed = 115200
extra_scripts = 
	pre:get_version.py
```

Basically any additional variants of a given board's firmware can be created simply by adding new entries to platformio.ini.

neilenns:
@elral would you mind taking a peek at this to ensure it looks good then I'll merge? Thanks!

elral (approved):

PR #45, opened 2021-09-22
Elral/issue36

elral:
In the function OnGetConfig() the EEPROM content is read bytewise and directly transfered via CMDMessenger. For this an additional function in the CMDMessenger lib is required which I named "sendSingleArg(T arg)". I added it in the for now used lib (in /lib) and in the forked lib which platformio.ini uses.
No further changes are required.
This pull request will superseed pull request #13.


danecreekphotography:
Of the two pull requests this is the one that seems the clearest to me. It's obvious that the code is reading the config and then just sending it out pre-formatted via CmdMessenger.

Will add comment over on the code side with a small suggestion.

lib/CmdMessenger/CmdMessenger.h
```diff
@@ -232,6 +232,17 @@ class CmdMessenger
       }
   }
   
+	/**
+	 * Send a single argument as string w/o field_separator
+	 *  Note that this will only succeed if a sendCmdStart has been issued first
+	 */
+	template < class T > void sendSingleArg(T arg)
```
danecreekphotography, line 239: Is there a better name for this? Maybe "sendRawByte"?

elral: Maybe just "sendArg()"?
All other functions habe "Cmd" and "Arg" in its name. This function shall send w/o a command, so just let it away.

danecreekphotography: How about `sendPreformattedArg()`? Then it follows the same naming pattern and is also really clear what it does.

elral: If a have a look on all send...() functions, they seem all to be preformatted. I think the comparible (existing) function is sendCmdArg(), but here w/o the command. So just as my previous proposal sendArg().
Maybe directly to add:
`void sendArg(T arg, unsigned int n);
void sendSciArg(double arg, unsigned int n = 6);
void sendBinArg(T arg);
`
like the other ones in the lib.

danecreekphotography: Works for me!

Issue #36, opened 2021-09-20
Remove duplicate copy of config buffer


danecreekphotography:
From https://github.com/danecreekphotography/MobiFlight-FirmwareSource/pull/13#issuecomment-921924372.

Currently a copy of the config buffer is made because strtok_r replaces the delimiter in the original string with a null character. @elral has a proposed solution that will remove this duplicate buffer copy.

elral:
The idea is only to keep the config once, process this config (replacing the delimiter with NULL) so that all pointers from the devices get there information and do not keep a copy with the original config. This duplicate config is "only" needed if the connectors requires the config vai `void OnGetConfig()`. Instead of sending the duplicate buffer this function now sends directly the content from the EEPROM byte wise. But this would mean also that the CMDMessenger has to be extended by a function where bytes can be send w/o field separator. In my fork I have this already added, and the changes required for MF I will propose soon.

elral:
I have to branches with slightly different changes. Both have advantages and disadvantages to discuss. Should I for both make a pull request?
And it would be really good if some one could also test these changes.

danecreekphotography:
Open two pull requests, the build system will automatically produce the firmware and attach it as a comment to the PR so it's easy to download and test.

PR #46, opened 2021-09-22
Elral/issue36_2

elral:
This is an alternative pull request to #36.
In this case the config will be loaded again from the EEPROM into the already processed configBuffer (all "," and ";" are replaced by NULL), then it is transfered using already existing CMDMessenger function and at least the configBuffer will be processed again. Smaller changes in onResetBoard() and loadConfig() are required. They original functions are uncomment with additional comments. Advantage is that no changes in CMDMessenger is required, disadvantage is that a connected LCD will be switched ON / OFF every time the connector connects or requests the config.

As also changes in the CMDMessenger for wrapping the #defines are required, I would prefer to add the required function as described in the pull request #36 and to merge not this one.

This pull request will superseed pull request #13.

### DIFF
--- a/include/mobiflight.h
+++ b/include/mobiflight.h
@@ -93,7 +93,7 @@ void OnResetConfig();
 void OnSaveConfig();
 void OnActivateConfig();
 void _activateConfig();
-void readConfig(char * buffer);
+void readConfig();
 void OnUnknownCommand();
 void OnGetInfo();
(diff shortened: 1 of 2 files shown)


danecreekphotography:
@elral I think it's safe for you to close this PR, I assume you are also more inclined to go with your first option for fixing this.

PR #53, opened 2021-10-10
Functions of LedControl into MFSegemts, "Fixes #49" and "Fixes #52"

### LINKED ISSUE #49: ProMicro crashes after uploading a new configuration
This has to do with running out of memory. @elral investigated and reports this in Discord:

> Stepper, Segments and LCD lib allocates there required memory on runtime. So far so good. BUT after uploading a new config the memory must be freed (delete class), and this does not really work on the Arduinos (what I have read). Most worse in this aspect is the 7Segment lib, as this one uses a bigger array in some functions (which are not used for now in the firmware).

A fix was also explored and mentioned in Discord:

>  I used only the required functions from the 7Segment lib and implemented them directly into the MFSegment.cpp (I guess Sebastian is not very happy with this 😉 ) so that they are used statically (for what I have understood). With these changes the promicro does not crash anymore after uploading a new config. Furthermore I could save some more RAM usage with this change and with the already prepared changes it seems that the promicro runs stable with all devices activated. The only thing I haven't tested (just defined in the config) are the shift registers.

We need to decide if this is the approach to take to resolve this issue, then implement for real.

### LINKED ISSUE #52: Mega doesn't report back all configuration after reloading configuration
From Jamie:

> on the mega firmware. It was working very well, but just to test I reloaded the mfmc file for the board. The upload to the board appears to go well, but after exiting the MF Modules window, I was having problems and going back to MF Modules now showed the board is missing about half the configuration. I repeated this three times with the same result.

elral:
## Description of changes

Fixes #49 and #52

Required functions from LedControl lib are integrated into MFSegments.cpp. With these changes a buffer[64] can be avoided which will save additional RAM.
Furthermore no memory is allocated  on runtime `(new LedControl()`) and more important no `delete _ledControl` in case of writing a new config is required. `delete` seems not to be working properbly and fragmentates the memory.


neilenns:
> Can I ask that you change all the added from LedControl:: to have the driver name in it so that other drivers can be added.

This PR is scoped to just the work required to fix some memory issues. Let's hold off on any additional changes until a future pull request so we can get this work completed and out the door.

wotupfoo:
@elral I'm actually of the philosophy that MFSegment should be made generic. The only per-driver info would be an enum of the supported types. Then, just as the LedControl lib is, have the 16K33 and TM1637 in their own libraries.
If you strip down the 3 libraries into bare bones MF only, instead of general purpose libs, they should be very small.
It's why I cloned them to my repo - to gut them of the unused functions since link-time-optimization (-flto) doesn't always remove unused functions in a class, and isn't supported in every toolchain linker.
To make MFSegment even more generic you could potentially Template the 3 drivers so that no if/else/switch is needed to call the class members.

`/* An enumeration of the supported types of LED digit displays */
enum eSegmentType
{
    kSegmentType_undefined = 0,
    kSegmentType_7seg = 7,
    kSegmentType_14seg = 14,
    kSegmentType_16seg = 16,
    kSegmentType_16x8matrix = 128,
};

/* An enumeration of the supported multi-segment and dot-matrix drivers */
enum eSegmentDriver
{
    kSegmentDriver_undefined = 0,
    kSegmentDriver_MAX1719, // SPI
    kSegmentDriver_MAX1721, // SPI
    kSegmentDriver_HT16K33, // I2C
    kSegmentDriver_TM1637   // pseudo I2C
};`

PR #55, opened 2021-10-14
Remove EEPROMEx library

elral:
No issue for this pull request so far.

Up to now I havn't understood why the EEPROMEx lib is used. If you break down the used functions only fundamental EEPROM commands are used.
By using it in the proposed way some bytes of memory are saved. Main advantage is that only small changes in this one file would be required for further using other boards/processors.


neilenns:
A namespace would be another way to do it, then refer to EEPROM::init etc.

PR #79, opened 2021-11-03
Fix: corrected byte count in shift

### LINKED ISSUE #80: Byte count incorrect in shift-out
The function that sends data out to the shift registers is off by one in the byte count.
The behavior does not cause apparent malfunctions becaus the excess byte is discarded.
Fixed by #79.

GioCC:
## Description of changes

Fixes #80:
The count of bytes shifted out was incorrect. It did not affect operation because the excess byte was on the tail end and was discarded, however it unnecessarily increased timing and might cause problems in the future.

## Checklist

If your change touches anything under `src/` or the `README.md` file
these items must be done:

- [v] Add a user friendly description of the change to the `unreleased` section of `CHANGELOG.md`


neilenns:
Can you also open a matching issue for this? Thanks!

GioCC:
Sure. I hadn't thought about it because it had no apparent malfunction, but no problem.

neilenns:
It's always good to have an issue matching each pull request. Makes it easier to see the history of work in the project. Thanks!


neilenns:
The change makes sense to me. I tried testing it locally but unfortunately my test setup was having issues even with the old firmware. I'll try again tomorrow (I had this problem before and can't remember what was causing it. I suspect a bad wire on the breadboard).

neilenns:
Took me *forever* but I wired a board up with two 74HC595s and managed to get the wiring correct and confirmed this change works as expected.

neilenns (approved):

elral:
As I understand this pull request ,all changed files for the "handler" changes are still part of this one which are now in a separate pull request.

GioCC:
> As I understand this pull request ,all changed files for the "handler" changes are still part of this one which are now in a separate pull request.

You're right, and that's unintended! My bad, I noticed this after @DocMoebiuz pointed it out yesterday. I will have to be more careful next time, for now it's probably better to leave things this way (or otherwise suggestions are welcome on how to proceed).

neilenns:
@GioCC Since this is such a small change the easiest way to fix this is to just do it again. Close this pull request, delete your branch, make a new one, apply the change, and open a new PR.

There are other ways to fix this with `git revert` but it's honestly easier to just make a new branch/PR for a chance this size.

Make sure to fetch from upstream/main first and merge to your main branch so the new branch is up-to-date with recent changes in the upstream repo.

GioCC:
@neilenns , sensible advice. I will definitely do that this evening back from work.

PR #81, opened 2021-11-07
Optimization and simplification of Event handler management 

GioCC:
## Description of changes
(Improvement - no issue fixed)

_NOTICE: These modifications still need testing, but I would first like your feedback about the assumptions I made.
E.g.: once the callback pointers are made static, leaving in place different entries for different transitions might still be unnecessarily cluttering, but with no penalty in terms of memory usage.
One more point to think about: analog inputs are the only type of input which might be more likely to require a different handler depending on the source channel._

In order to optimize memory usage and simplify code logic,  the Event handler functions have been modified as follows:
* handlers for MFButton unified, also for MFEncoder (all transitions use the same handler anyway)
* handlers for MFButton, MFEncoder and MFAnalog (pointers and assignment functions) made static (no need to define a different handler for each button. Exception: Stepper zero, see remark below).
* MFAnalog: removed handler assignment from constructor; introduced separate attachHandler() function like MFButton/MFEncoder
* MFStepper: removed dependency from MFButton in order to avoid the need for separate handler (not a "proper" button anyway)
* (not related to the main topic) MFButton: added status argument to trigger() to avoid unnecessary double call

The changes are a little extensive as far as the number of files is concerned, but not really pervasive.


GioCC:
A little comparison of storage size, if that helps.
Before: 
```Mega
RAM:   51.0% (used 4178 bytes from 8192 bytes)
Flash:  9.8% (used 24832 bytes from 253952 bytes)

ProMicro
RAM:   67.3% (used 1723 bytes from 2560 bytes)
Flash: 88.7% (used 25432 bytes from 28672 bytes)

Uno
RAM:   73.3% (used 1502 bytes from 2048 bytes)
Flash: 71.3% (used 23012 bytes from 32256 bytes)
```
After:
```Mega
RAM:   45.3% (used 3714 bytes from 8192 bytes)
Flash:  9.5% (used 24094 bytes from 253952 bytes)

ProMicro
RAM:   63.2% (used 1617 bytes from 2560 bytes)
Flash: 86.1% (used 24694 bytes from 28672 bytes)

Uno
RAM:   70.7% (used 1448 bytes from 2048 bytes)
Flash: 69.1% (used 22290 bytes from 32256 bytes)
```

elral:
While setting up my first draft for a keymatrix I also struggled about all these handlers. From my point of view this makes absolutely sense, I can do tests earliest on the weekend, maybe in the next days for a promicro but w/o device connected.
Just a few comments from my side:
* Can the handlers for buttons/encoders/analogIn combined to one? They all transfer a "value" and name. The only difference is uint8_t / uint16_t (analog).
* Must it be in this way like the handlers are used now or would be a global "function" (which would be called from each device) also sensefull/easier? Yes, I am not the best in C++ ;)

elral:
I am wondering what the best strategy would be to test these changes. I could set up a branch in my fork, but how do I get all these changes into my branch? Or can I directly "copy" a branch from gioCC into my repo?

GioCC:
My take on your welcome comments:
> * Can the handlers for buttons/encoders/analogIn combined to one? They all transfer a "value" and name. The only difference is uint8_t / uint16_t (analog).

They probably could, though I am not sure if that would bring a practical advantage. The way I see it, keeping it this way leaves  a minimum degree of flexibility that could be useful.

> * Must it be in this way like the handlers are used now or would be a global "function" (which would be called from each device) also sensefull/easier? Yes, I am not the best in C++ ;)

Similar answer as above: probably in this case an "ordinary" global function would do quite fine, but then again, this way the code has a little more sound decoupling at basically negligible cost.

GioCC:
> I am wondering what the best strategy would be to test these changes. I could set up a branch in my fork, but how do I get all these changes into my branch? Or can I directly "copy" a branch from gioCC into my repo? I would appreciate some hints.

I guess the safest and more straightforward thing would be to clone my repository in a new folder. Probably a little early to think about a merge (even temporary and for testing) with other code.

neilenns:
@elral VSCode makes it easy to review/test pull requests.

1. Make sure you have the [GitHub Pull Requests and Issues extension ](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) installed.
2. Click the GitHub icon in the left icon rail
3. Right click on the pull request in the top part of the screen and select _Checkout pull request_

image

You can also use the pull request UI in VSCode to review and comment on pull requests.

src/MF_Modules/MFButton.cpp
```diff
@@ -17,31 +19,30 @@ void MFButton::update()
     uint8_t newState = (uint8_t) digitalRead(_pin);
     if (newState!=_state) {     
       _state = newState;
-      trigger();      
+      trigger(_state);
```
neilenns, line 22: Why is a trigger method, which is just a wrapper around an if statement, which itself is just a wrapper around an event handler, which itself is just going to call cmdMessenger.sendCmd even required?

Why can't all those levels of indirection be removed and `update()` simply makes the calls to cmdMessenger directly?

neilenns (dismissed):
This is a cool start but can it be radically simplified to remove event handlers entirely?

src/MF_Modules/MFButton.cpp
GioCC: That is a question I would have gladly asked myself...!
I didn't want to push changes too far beyond what was really effective, because I didn't want to look too subversive :)
But I agree, in my opinion (see my previous reply to @elral) I would leave just the callback mechanism so as not to embed the details of cmdMessenger in this class. Unless @DocMoebiuz has anything against, it I'd be happy to go that way.

GioCC: Besides, event handlers for outgoing event messages mirror those required by cmdMessenger for incoming commands, which somehow makes sense from a logic standpoint at least.

neilenns: Ah, I see, that makes sense. Yes, so the individual MF things don't actually know about how the data is sent/received, and in the future some other transport mechanism could be used. Of course :)

elral: Yes, that makes sense.

GioCC:
> @GioCC , did you test the changes on a board? I just tried it and the one button and one encoder which is connected do not inform about changes.

Unfortunately not yet, I finished late yesterday and I will be able to make time not sooner than tomorrow evening.

GioCC:
I found out that I was one local commit short when I made the pull request.
After line 238 in mobiflight.cpp there should be another line "attachEventCallbacks();"
Sorry for the mishap, I will correct is as soon.

elral:
Just saw this while looking through the code. Now the button and encoder is working ;)

MobiFlight-Admin:
The original motivation for individual callbacks/handlers comes certainly from my OOP background. Now that we know that we don't need the flexibility (other than the stepper), and we want to lower the memory footprint, such a change can be made.

I am still not decided if i really like the MFButton removal... I think it is like a button, and we should reuse our own building blocks. So I am not sure if it is worth it to save a couple of bytes for that. but I see exactly this is the "blocker" for the static button eventhandler... because otherwise we would have to put extra logic there... which won't work.

GioCC:
@DocMoebiuz exactly, you've summarized it perfectly. I was also reluctant to remove the MFButton from the stepper at first, and I would have probably left it if it hadn't blocked the whole modification... but in the end, even conceptually the stepper switch is unlikely to require most of the features of a generic button, and only a few lines of code are actually required. After all, I was convinced the choice was sound.

elral:
For me the stepper switch is a limit-/position switch, not an input switch. The function is different, the position switch should detect a defined position and this without debouncing. The first signal change must be detected and reacted, following debouncing is not really reproducable (and therefore the time its debouncing).
Input switches must detect exactly one signal change if pressed from the user, so they must be debounced. 2digit milliseconds later (debounced state) from the first signal change do not care.

But there is another topic I can't answer myself. I think in the existing version a signal change is detected and a message is given to the connector via the MFButton class. I don't know if this is processed by the connector.
With this new version I don not see that a message is given to the connector. If it would be required it could be added. Unfortenutely I do not know the connector in detail to answer this.

GioCC:
> I think in the existing version a signal change is detected and a message is given to the connector via the MFButton class. I don't know if this is processed by the connector. With this new version I don not see that a message is given to the connector. If it would be required it could be added. Unfortenutely I do not know the connector in detail to answer this.
That is another reason why I considered the stepper switch not to be a "proper" button. 
I didn't follow up on the check inside Connector, but the MFButton belonging to the stepper was not registered (thus it did not took part in forced retriggers), nor would transmit any status change (its callback had no calls to cmdMessenger).

MobiFlight-Admin (approved):

GioCC:
@MobiFlight-Admin @neilenns I have one question to ask: I had already removed myself `MF-FirmwareSource_GCC.code-workspace` and `nuovo 1.txt` myself from my branch (and committed to the remote repo); I would have expected the deletions to appear in the PR, did they not? Do you have any suggestion about what I might have missed?

neilenns:
They should have appeared as long as you did these three steps:

1. Delete the files
2. Commit the changes
3. Push to the remote

It's possible you did 2 but not 3.

GioCC:
I thought I did all three, but it's definitely possible I'm mistaken. I'll take better care next time.

PR #83, opened 2021-11-07
ProMicro filename is wrong

### LINKED ISSUE #82: ProMicro filename is wrong
It's currently "promicro" and should be just "micro" for backwards compatibility with the old firmware process.

neilenns:
Fixes #82

## Description of changes

Change the name of the platform in platformio.ini so the firmware has the correct name


MobiFlight-Admin (approved):
Thanks!

PR #91, opened 2021-11-09
Fix: corrected byte count in shift (NEW PR)

### LINKED ISSUE #80: Byte count incorrect in shift-out

GioCC:
(Replaces PR #79)
Fixes  #80:
The count of bytes shifted out was incorrect. It did not affect operation because the excess byte was on the tail end and was discarded, however it unnecessarily increased timing and might cause problems in the future.


neilenns:
I tested this change with two chained shift registers and confirmed they still work properly.

MobiFlight-Admin (approved):

PR #96, opened 2021-11-13
Support any combination of row/column counts for LCD displays

### LINKED ISSUE #85: MFLCDDisplay::display() allocates a fixed array of 21 chars. This fails with 40 char LCDs
The buffer overflow with 40 chars corrupts the class storage, which leads to unpredictable behaviour.

I've rebuilt the Firmware with a new MAX_COL set to 40, uploaded to the Arduino controlling a 40 x 2 display. Works fine now.

Ideally ::attach() would return bool or byte, with an explicit error code for "too many columns". Then
the GUI can take appropriate action.

elral:
Fixes #85

Instead of copying the single line to a buffer, print directly from the char buffer containing the complete message via ->write().

Tested with a 20x4 LCD, but will work with any combination of columns and rows.

### DIFF
--- a/src/MF_Modules/MFLCDDisplay.cpp
+++ b/src/MF_Modules/MFLCDDisplay.cpp
@@ -18,7 +18,10 @@ void MFLCDDisplay::display(const char *string)
     _lcdDisplay->setCursor(0, line);
     for (uint8_t col = 0; col < _cols; col++)
     {
-      _lcdDisplay->write(string[(_cols * line) + col]);
+      uint8_t char2print = string[(_cols * line) + col];
+      if (!char2print)                                     // just to be sure not to print after NULL termination
+        return;
+      _lcdDisplay->write(char2print);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


MobiFlight-Admin (commented):
@elral nice fix - we should think of checking whether we have reached the end of the string to not go beyond it if the string was shorter. Just in case. What do you think?

MobiFlight-Admin:
@elral nice fix - we should think of checking whether we have reached the end of the string to not go beyond it if the string was shorter. Just in case. What do you think?

elral:
@MobiFlight-Admin yes, good idea. I will change the pull request accordingly.

MobiFlight-Admin (approved):

Gamma36:
Does writing single characters slow down the LCD update?
Just asking as a code check.

Thanks for the fix

elral:
No, it does not slow down. Print() does the same, writing single characters.
And doing it this way you don't need to copy the memory area.
In the end it is slightly faster.

PR #106, opened 2021-11-30
Changes for modified LiquidCrystal_I2C lib

elral:
## Description of changes
With pull request https://github.com/MobiFlight/LiquidCrystal_I2C/pull/1 the parameters are moved from the constructor to the init() function. Therefore this has to be changed also in the firmware.


MobiFlight-Admin (approved):
THANKS!

PR #114, opened 2021-12-03
Read time for buttons and analog in is wrong Fixes #113

### LINKED ISSUE #113: Read time for buttons and analog in is wrong
The intention for reading the buttons was to do it every 10ms. With the current implementation buttons are read every 11ms.
Same for analog reading, intentation was to calculate the average every 10ms, but is 11ms.

In the setup() function an offset for the actual millis() is defined to decouple the read in of the input devices. With 1 ms more it will come to regulary overlap of the read in functions.

elral:
Read in time for buttons reduced from 11ms to 10ms as originally intended
Read in time for analog average reduced from 11ms to 10ms as originally intended

Fixes #113

### DIFF
--- a/include/mobiflight.h
+++ b/include/mobiflight.h
@@ -5,6 +5,8 @@
 #include <MFAnalog.h>
 
 #define MF_BUTTON_DEBOUNCE_MS   10      // time between updating the buttons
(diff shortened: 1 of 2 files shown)


MobiFlight-Admin (approved):
Thanks

PR #118, opened 2021-12-08
Mobiflight FW modifications for multiplexer management

GioCC:
- _Subject_: implementation of hardware multiplexer management
  - __General multiplexer driver__ - can be readily used to support ___multiplexed routing of existing I/O blocks___ (extending them with very minor modifications)
  - __Digital input block__ (comparable to shift registers in function) using MPX ICs

- _Status_: __Alpha stage__ - code modifications complete, firmware only (no Connector), compilation OK, untested.  
  PR posted as draft for early review.

Refers to issue  #119 
  
## Class MFMultiplexer

- _Purpose_: manage 4 selection lines for external multiplexers, on any Arduino pins (defined as type `kTypeMultiplexer`).  
  These lines will be connected in parallel to every hardware multiplexer used (typically 74HCT4067 or -4051).
- One instance only (variable `MPX`), activated with following command:
 
    `kTypeMultiplexer.Sel0_pin.Sel1_pin.Sel2_pin.Sel3_pin`

- The multiplexer can be used alone (to route inputs or outputs of other blocks such as display drivers, shift registers etc), or as required component for other blocks (e.g. direct multiplexed digital / analog inputs).  
  - In the first case, I/O blocks will exploit the fact that the MPX automatically increments the selector value at every iteration of the main loop;  
  blocks can read the current selector value and determine their behavior accordingly.
  - In the second case, blocks can drive the selector as they require (e.g. doing a complete scan) with no interference with the main loop iteration.

## Class MFMPXDigitalIn

- _Purpose_: manage 8/16 digital input lines, multiplexed on a single Arduino pins (defined as type `kTypeMPXDigitalIn`). These lines will be connected to a hardware multiplexer (typically 74HCT4067 or -4051), whose selector lines will be connected to the pins defined by the MFMultiplexer component defined separately.
- Several instances (see  `MAX_DIG_IN_MPX` in files `MFBoards.h`), activated with following command:  

    `kTypeMPXDigitalIn.dataPin.halfSize.mode.name`  

  where:  
  - `dataPin` is the common Arduino input pin
  - `halfSize` is `0` for 16-channel MPXs (eg 4067) or `1` for 8-channel MPXs (eg 4051)
  - `mode` is `0` if the block should perform a full scan whenever polled, or `1` if it should read on bit only for each polling cycle (following the value of the main loop MPX count).  
  In both cases, changed inputs immediately generate event callbacks (of type `kMPXDigitalInChange`).


neilenns:
I took a quick peek at this in the context of my pending input shifter support. There's a lot of similarities between MPXDigitalIn and the way I did input shifter support. Could I somehow use MPXDigitalIn?

If not, the way the `detectChanges()` code is written should align so it looks like we took the same approach. Right now I think we both wrote code to do the same thing but the way we wrote it is very different.

GioCC:
@neilenns  TBH, there's a reason for that... I took your InputShifter code as a starting template, so as to be sure I did not forget some piece around :D

The difference between the two blocks should reside mainly in the `poll()` method (and the fact that you can have more then two SR chained), otherwise they do pretty much the same things. I'm not sure whether it would be beneficial to try and join the code in a common class (or e.g. an ancestor class to inherit from), if that's what you mean, but I'd like to hear your point of view.

The only different thing I wanted to suggest for porting to the Input Shifter code would be the way the retrigger is handled, by exploiting `detectChanges()` explicitly (and modifying `poll()` to refrain from calling it).

BTW, I would be very grateful if you could find some time for a little help with setting up the Connector for testing, since I understand you've already worked on it recently; I have no problem working with C#, but I'm no longer familiar with the Mobiflight application at all.

GioCC:
Replaced with a new, updated PR.

PR #121, opened 2021-12-11
After start first step of enc is always fast step

### LINKED ISSUE #120: After startup the first step of an encoder is always a fast step
Sebastian recognized, that after an initial boot the first detent of an encoder is always a fast detection.

Reason is, that _positionTimePrev and _positionTime is not initialized in  MFEncoder::attach() and therefore both have the same value. This results in a high speed calculation.

elral:
Fixes #120
First step of an encoder after startup is always a fast step. This is due to missing initialization of _positionTimePrev and _positionTime. Therefore they have the same value and a high speed is calculated.


MobiFlight-Admin (approved):
Thanks!!!

PR #116, opened 2021-12-07
Optimization of unused code in command parsing

GioCC:
## Description of changes

Fixes #115 (code cleanup / improvement)
Prevents compiling unused chunks of code in command parsing.

PR #122, opened 2021-12-20
Optimization of unused code in command parsing [resubmitted]

### LINKED ISSUE #115: Optimization of unused code in command parsing readConfig()
Not a big deal, but some code can be left off according to compilation options. (Will update with reference to PR).

GioCC:
## Description of changes

Fixes #115
Replaces PR #116 (tied to wrong branch)

### DIFF
--- a/src/mobiflight.cpp
+++ b/src/mobiflight.cpp
@@ -770,6 +770,7 @@ void readConfig()
       break;
 
     case kTypeLedSegment:
+#if MF_SEGMENT_SUPPORT == 1
       params[0] = strtok_r(NULL, ".", &p); // pin Data
       params[1] = strtok_r(NULL, ".", &p); // pin Cs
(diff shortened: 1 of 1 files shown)


DocMoebiuz (commented):
Thanks for this proposal.

I just wondered: It would make sense to enclose the entire "case"-branch with the #if #endif, don't you think?

We would not have to deal with the same
```
#else
      strtok_r(NULL, ":", &p); // skip unmanaged command
#endif
```
because that would be done by the default case then.

DocMoebiuz (changes requested):
see my last comment.

GioCC:
Yes, of course it makes sense. Silly me :) 
I'll change it right away.

DocMoebiuz (approved):
YEAH!

PR #124, opened 2021-12-23
Config Buffer gets filled with unnecessary bytes

### LINKED ISSUE #123: Config Buffer gets filled with unnecessary bytes
During a config upload from the connector the complete config buffer gets filled with not required bytes. The end of the config is still detected by 0x00, but it is better not doing so and it decreases the upload time.

Sebastian has it already implemented for the upcoming version of the FW.

elral:
Fixes #123

configBuffer was on every part of the upload written up to the end of the buffersize. Every byte after the config part came from "somewhere".

With this change only the received part of the config  gets copied.

### DIFF
--- a/src/mobiflight.cpp
+++ b/src/mobiflight.cpp
@@ -673,7 +673,7 @@ void OnSetConfig()
 
   if (configLength + cfgLen + 1 < MEM_LEN_CONFIG)
   {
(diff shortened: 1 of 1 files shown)


MobiFlight-Admin (approved):
danke

PR #129, opened 2022-01-04
Use progmem for two fixed strings

### LINKED ISSUE #128: Move string constants to PROGMEM
There are a couple of strings in the code that are constants (board type, version) but aren't stored in PROGMEM. These are easily movable by using the F() macro.

neilenns:
Fixes #128

Move two strings to PROGMEM. On a ProMicro before this change:

```
RAM:   [========  ]  77.5% (used 1984 bytes from 2560 bytes)
Flash: [========= ]  91.6% (used 26254 bytes from 28672 bytes)
```

On a ProMicro after this change:

```
RAM:   [========  ]  76.6% (used 1962 bytes from 2560 bytes)
Flash: [========= ]  91.7% (used 26290 bytes from 28672 bytes)
```

Saves 22 bytes of RAM at the expense of 36 bytes of flash.

### DIFF
--- a/src/mobiflight.cpp
+++ b/src/mobiflight.cpp
@@ -61,11 +61,10 @@ const uint8_t MEM_OFFSET_CONFIG = MEM_OFFSET_NAME + MEM_LEN_NAME + MEM_LEN_SERIA
 
 uint32_t lastAnalogAverage = 0;
 uint32_t lastAnalogRead = 0;
-uint32_t lastButtonUpdate= 0;
+uint32_t lastButtonUpdate = 0;
 uint32_t lastEncoderUpdate = 0;
 uint32_t lastServoUpdate = 0;
 
-const char type[sizeof(MOBIFLIGHT_TYPE)] = MOBIFLIGHT_TYPE;
 char serial[MEM_LEN_SERIAL] = MOBIFLIGHT_SERIAL;
 char name[MEM_LEN_NAME] = MOBIFLIGHT_NAME;
 const int MEM_LEN_CONFIG = MEMLEN_CONFIG;
@@ -183,9 +182,7 @@ void attachEventCallbacks()
 #if MF_ANALOG_SUPPORT == 1
   MFAnalog::attachHandler(handlerOnAnalogChange);
 #endif
-
 }
-
 
 void OnResetBoard()
 {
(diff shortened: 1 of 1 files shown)


src/mobiflight.cpp
```diff
@@ -61,11 +61,10 @@ const uint8_t MEM_OFFSET_CONFIG = MEM_OFFSET_NAME + MEM_LEN_NAME + MEM_LEN_SERIA
 
 uint32_t lastAnalogAverage = 0;
 uint32_t lastAnalogRead = 0;
-uint32_t lastButtonUpdate= 0;
+uint32_t lastButtonUpdate = 0;
 uint32_t lastEncoderUpdate = 0;
 uint32_t lastServoUpdate = 0;
 
-const char type[sizeof(MOBIFLIGHT_TYPE)] = MOBIFLIGHT_TYPE;
```
neilenns, line 68: This is no longer needed since the string comes from PROGMEM.

src/mobiflight.cpp
```diff
@@ -884,10 +883,10 @@ void OnGetInfo()
 {
   lastCommand = millis();
   cmdMessenger.sendCmdStart(kInfo);
-  cmdMessenger.sendCmdArg(type);
```
neilenns, line 887: Use the `F()` macro to get the string to come from PROGMEM.

src/mobiflight.cpp
```diff
@@ -884,10 +883,10 @@ void OnGetInfo()
...
+  cmdMessenger.sendCmdArg(F(MOBIFLIGHT_TYPE));
   cmdMessenger.sendCmdArg(name);
   cmdMessenger.sendCmdArg(serial);
-  cmdMessenger.sendCmdArg(VERSION);
```
neilenns, line 890: Use the `F()` macro to get the string to come from PROGMEM.

neilenns:
Note that there are a bunch of whitespace modifications as well due to VSCode autoformatting.

DocMoebiuz (approved):
Many changes due to re-formatting

Thanks for the PROGMEM improvement

DocMoebiuz (approved):

MobiFlight-Admin (approved):
thanks

PR #133, opened 2022-01-07
Add missing #if for analog and servo

### LINKED ISSUE #130: Analog-specific variables aren't omitted if analog is disabled
The variables used for last analog read/last analog average aren't ifdef'ed like the rest of the analog code. In addition some of the updates to the last read variables aren't ifdef'ed either.

https://github.com/MobiFlight/MobiFlight-FirmwareSource/blob/3d7bc58a3a860e312d6dc1d433683ca220096bd8/src/mobiflight.cpp#L62-L63

https://github.com/MobiFlight/MobiFlight-FirmwareSource/blob/3d7bc58a3a860e312d6dc1d433683ca220096bd8/src/mobiflight.cpp#L207-L211

neilenns:
Fixes #130

* Add missing #if statements for the analog and servo variables.

* Re-order the two places where those variables are mentioned to be consistent.


neilenns:
Here's the flash savings after the change when the analog and servos are disabled via the `MFBoard.h` file:

image

DocMoebiuz (approved):
THANK YOU!

Issue #138, opened 2022-01-11
Change internal data structure of (output) Shifters


GioCC:
For consistency with InputShifters; use a byte array in place of a fixed word.
No practical advantages currently.

neilenns:
Agreed, it always bugged me that it wasn't done using arrays.

GioCC:
It has a rationale, though, because exceeding 32 bits is rather unlikely; that would mean a longer hardware chain and that's not something you would want to have. But hey, it bugged me a little too.

Issue #137, opened 2022-01-11
Update "Shifter" naming (as "OutputShifter") for better consistency


GioCC:
Just a cosmetic measure, but avoids confusion with InputShifters (and looks cleaner).
Involves several names inside the code, but it's just a matter regarding the source code with no functional impact.

PR #139, opened 2022-01-11
Output Shifter upgrade (naming and data structure)

GioCC:
## Description of changes

Addresses issues #137 (first commit) and #138 (second commit)

The naming upgrade only affects source code, no difference in functions or performance.

The data structure changes the fixed-length buffer with a byte array; there is a slight penalty in Flash space (+18 bytes), though - surprisingly - if we dispense with one single boundary check for an argument (in setPin()) there is even a slight gain (-12 bytes)!

_The second part still needs testing_


src/MF_Modules/MFOutputShifter.cpp
```diff
@@ -0,0 +1,116 @@
...
+    updateShiftRegister();
+    delay(50);
+  }
+  if(mode==0) {
+    _outputBuffer[_moduleCount-1] = 0;
+    updateShiftRegister();
+  }
+}
+
+void MFOutputShifter::test()
```
neilenns, line 95: This change is unrelated to the purpose of this PR and doesn't have a tracking issue. Did you mean to include it?

neilenns (changes requested):

src/MF_Modules/MFOutputShifter.cpp
GioCC: Not really. I was trying to introduce a maybe more useful, possibly less "expensive" test mode, when I discovered that the `test()` method is not invoked anywhere, so I forgot to follow up on that.

neilenns: Can you push an update that removes it then? Thanks!

GioCC:
Done!

src/MF_Modules/MFOutputShifter.cpp
```diff
@@ -0,0 +1,94 @@
...
+// Copyright (C) 2021
+
+#include "MFOutputShifter.h"
+
+MFOutputShifter::MFOutputShifter()
+{
+  _initialized = false;
+}
+
+void MFOutputShifter::setPin(uint8_t pin, uint8_t value, uint8_t refresh)
```
neilenns, line 12: What is this "refresh" thing? It doesn't get passed in anywhere that I can see and is unrelated to the naming changes.

neilenns (changes requested):

src/MF_Modules/MFOutputShifter.cpp
GioCC: It is used in SetPins(); it allows to change the output values deferring the transmission once all changes have been done. This way, SetPins() can use SetPin() (instead of duplicating the code) because it otherwise it would send all data every time.

neilenns: Thanks for the explanation! This change wasn't listed in the PR description so it wasn't clear if it was intentional as part of the rename work (since it's unrelated to the two mentioned issues).

neilenns (approved):

src/MF_Modules/MFOutputShifter.cpp
GioCC: Actually I saw it more as part of the update related to the changes in the data structure, even if not strictly. It's internal management more than anything else, but I see it's clearer now. Thanks.

DocMoebiuz (approved):
Thanks!

PR #141, opened 2022-01-14
All boards can crash after upload w/ 4 7Segment

### LINKED ISSUE #140: Mega/ProMicro/Uno can crash after upload if 4 7Segments are configured
Once 4 7Segments are configured and an new config upload is done, all boards can crash. With an own Buffer for the device they will crash.

Before the new config will be uploaded, all devices are cleared. Within `ClearLedSegments()` the for loop shall clear all configured devices, but instead the loop variable `ledSegmentsRegistered` is used and there the buffer size is exceeded.

Solution is to exchange `ledSegmentsRegistered` by `i`.

elral:
Fixes #140

Eexchange ledSegmentsRegistered by i in function ClearLedSegments().
Otherwise the ledSegments[] buffersize will be exceeded when 4 7Segments are configured


neilenns:
Nice find!

elral:
Unfortunetly I missed one more buffer overflow, but now I should have find all of them.

First tests shows that it runs very stable now.

elral:
Sorry, I forgot to mention, that this is not only an issue regarding a possible buffer overflow. This "only" happens if the max. number of a device is configured. Much more important is, the memory wasn't cleared up after a config upload if one or more devices from these 6 devices were configured.

But I still recommended to merge issue 127. The pull request is nearly done and more explanations will be icluded in the upcoming pull request.

DocMoebiuz (approved):
LOL!

Thanks for finding and fixing

PR #145, opened 2022-01-20
Move common platform.ini pieces to a shared environment block

### LINKED ISSUE #144: Move common platform.ini pieces to a shared environment block
Right now most of the settings across the different board types are repeated in the platformio.ini file. These can instead use the common `[env]` section and then get referenced in the board-specific sections.

neilenns:
Fixes #144

Created common `env` sections for the parts of the build definition that are common across all boards.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -16,77 +16,74 @@
 ; by clicking on the target name in the bottom status bar.
 [platformio]
 
+; Common build settings across all devices
+[env]
+lib_deps = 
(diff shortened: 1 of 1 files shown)


elral (approved):
Great! There is so much in platformIO I just heared or even do not know ;)

GioCC (approved):

DocMoebiuz (approved):
THANKS!

PR #146, opened 2022-01-20
Add support for Raspberry Pi Pico builds

neilenns:
This is a draft pull request to get feedback on. @elral did all the base work to get this to run on Pico, including writing the eeprom methods and figuring out the various packages and platforms required to get it to do a base build in PlatformIO on Windows.

With his core work I've done the additional steps to get it working in our automated build process:

- Changed how SDA/SDL pins are assigned for Pico boards, using the [pin reassignment methods available in the framework](https://arduino-pico.readthedocs.io/en/latest/pins.html).
- Updated the PlatformIO build process to support both Windows and Linux build platforms due to the dependency on a platform-specific toolchain. Unfortunately there's no way around this until `maxgerhardt/toolchain-pico` gets [published properly to PlatformIO](https://github.com/earlephilhower/pico-quick-toolchain/issues/5).
- Updated the pre-build script to strip `_linux` off the Pico environment name so the firmware filename looks right.
- Updated the GitHub build scripts to build using the Linux version of the Pico toolchain.

The worst part of all this are the changes to platformio.ini. It required explicitly defining `default_envs` which means the build process no longer Just Works any time new board types are added. We'll always have to remember to go edit the build scripts as well when adding new boards.

I spent several hours trying to find a better way to do the linux/windows split and came up empty. There's no conditionals available in `platformio.ini` and I couldn't find any way to modify the `platform_packages` property via pre-build Python. Sad.


neilenns:
Note that the firmware.zip that auto-builds with this PR won't include the Pico firmware since the build script changes aren't merged.

DocMoebiuz:
great team work guys - i am really blown away by all your cool ideas and efforts you're putting in!!! Thanks a lot!

elral (approved):

neilenns:
Looks like I might have to disable caching on the github builds.

elral:
Since the last pull requests merged into main the standard [env] in platformio.ini was added. Part of the [env] lib_deps is `arduino-libraries/Servo @ 1.1.8` which will lead to a build error. As this lib is not required for the Pico it has to be moved back to the boards envirements.

neilenns:
> Since the last pull requests merged into main the standard [env] in platformio.ini was added. Part of the [env] lib_deps is `arduino-libraries/Servo @ 1.1.8` which will lead to a build error. As this lib is not required for the Pico it has to be moved back to the boards envirements.

The Pico build uses a specific list of libraries, it doesn't rely on the global one.

PR #147, opened 2022-01-24
Elral/issue127

### LINKED ISSUE #127: Uno and ProMicro crashes after config upload once a stepper is defined
Once a stepper is defined the Uno and ProMicro crashes after the next or overnext (depends on how many additional devices are defined) config upload. On the Mega it needs ~16 uploads w/o reset in between, so it will be very seldom.

To reproduce it on the ProMicro/Uno, define a stepper and e.g. a button. Upload this config and add an additional device. Upload this config again. Does the uplaod end w/o error, at least the next coming upload will not be successfull.

The reason is, steppers are initialized dynamically. During config upload all devices are resetted and for dynamically initialized devices the used memory (RAM) is freed. Unfortenutley this does not really work like we also saw for the 7Segements before.

Initializing the steppers statically will increase the required memory up to ~98%, but it seems to work stable until you define one LCD. The LCD device is also intialized dynamically so we are running into the same problem. The remaining RAM might not be sufficient to initialize the LCD statically, but needs some more tests.

The only solutions I can imagine for now is to free up the required RAM by e.g. dispensing the config buffer, reduce the number of devices or reduce the size of the config buffer. 
Dispensing the config buffer would mean some (bigger?) changes on the connector side, but is the most effective for freeing up the RAM.

Any other proposels are welcome.

### LINKED ISSUE #140: Mega/ProMicro/Uno can crash after upload if 4 7Segments are configured

elral:
Fixes #127 and fixes #140

These are modifications to save as much RAM as possible (I think so), increase stability and have full control on required memory for all devives which are now initialized dynamically.

Following changes are implemented:
* used pins not checked anymore (delete pinsRegistered)
* configBuffer contains only names from input devices
* MFEEPROM function getLength returns the size of the EEPROM
* Issue #140
* analogwrite changed to function from class
* debug printout to connector (should we stay with this approach?)
* outputs, buttons, encoders, anlogIn, OutputShifter and InputShifter dynamically _initialized
* MFStepper dynamically initialized in MFStepper.cpp like other devices which use a library
* if config for steppers are (pin2 == pin4 && pin1 == pin3), stepper lib will be initialized for external drivers (proposal, would require changes in the connector)
* MFLCDDisplay dynamically initialized
* MFSegments dynamically initialized
* MFServo dynamically initialized

This results in following memory saving:
image

The memory consumption for the device is as above mentioned:
* Encoder: 31 Byte
* Button: 4 Byte
* 7Segment: 24 Byte
* Stepper: 68 Byte
* Servo: 3 Byte
* LCD: 14 Byte
* Analog: 29 Byte
* Output: 2 Byte
* Output Shift Register: 9 Byte
* Input Shift Register: 11 Byte

Worst Case calculation for deviceBuffer:
image

So considering the available pins per board every configuration should fit into the new deviceBuffer.


GioCC (approved):

DocMoebiuz (approved):
Woohoo!!!! THANKS FOR THE EFFORT - YOU ARE A GREAT TEAM!!!

PR #149, opened 2022-01-31
Issue642part display blanking

GioCC:
## Description of changes

Implements part of #642: now setting Brightness = 0 completely switches the display off rather than setting minimum brightness
Memory occupation has been very slightly reduced by using the value of "_moduleCount" to replace the "_initialized" flag


GioCC:
Incidentally started from the wrong branch. Will reopen.

PR #150, opened 2022-01-31
Issue #151 - Brightness=0 completely switches display off

GioCC:
## Description of changes
Implements part of #642: now setting Brightness = 0 completely switches the display off rather than setting minimum brightness.
Actual value sent to the IC is lower by 1 (actual range: 0-14), since value 15 is effectively indistinguishable from 14, but value 0 is a very faint brightness that might be desired in some cases.
Memory occupation has been very slightly reduced by using the value of "_moduleCount" to replace the "_initialized" flag

### DIFF
--- a/include/MFSegments.h
+++ b/include/MFSegments.h
@@ -32,8 +32,7 @@ public:
     void setBrightness(byte module, byte value);
 
 private:
-    LedControl _ledControl;
-    //bool _initialized;
-    byte _moduleCount;
+    LedControl  _ledControl;
+    byte        _moduleCount;
 };
 #endif
--- a/include/MFSegments.h
+++ b/include/MFSegments.h
@@ -33,7 +33,7 @@ public:
 
 private:
     LedControl _ledControl;
-    bool _initialized;
+    //bool _initialized;
     byte _moduleCount;
 };
 #endif
--- a/src/MF_Modules/MFSegments.cpp
+++ b/src/MF_Modules/MFSegments.cpp
@@ -4,8 +4,6 @@
 
 #include "MFSegments.h"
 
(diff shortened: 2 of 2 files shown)


include/MFSegments.h
```diff
@@ -33,7 +33,7 @@ class MFSegments
 
 private:
     LedControl _ledControl;
-    bool _initialized;
+    //bool _initialized;
```
DocMoebiuz, line 36: Please remove the line instead of commenting it

src/MF_Modules/MFSegments.cpp
```diff
@@ -4,16 +4,17 @@
 
 #include "MFSegments.h"
 
+#define _initialized  (_moduleCount != 0)
```
DocMoebiuz, line 7: interesting, i personally find it is not as obvious to read in the code later with this inline replacement ... is there a convention in terms of UPPERCASE or similar? 

I would rather replace it with _moduleCount != 0 in the  places where we are using it or have a function that we call which will be replaced inline by the compiler later.

DocMoebiuz (changes requested):
Thanks @GioCC  - please see my comments.

include/MFSegments.h
GioCC: Oopsie.. forgot that.

src/MF_Modules/MFSegments.cpp
GioCC: You're right, capitalized would be better. And given it's small, even direct substitution would be OK.
An inline would probably be overkill.

DocMoebiuz (approved):
Thanks!

PR #152, opened 2022-02-08
Elral/issue148

### LINKED ISSUE #148: Split mobiflight.cpp into multiple files
@MobiFlight-Admin @neilenns @GioCC 
During creating the last pull request it was some kind of "difficult" to keep the overview in mobiflight.cpp. A lot of scrolling through the file was required to find all related parts of the code.
PlatformIO has better capabilities to use multiple files than the Arduino IDE, and discussions in the past shows that all of us agree to split mobiflight.cpp into severall files.
@neilenns has done this already by setting up an additional file where all devices and other parts are set up in a class, I did it in the past not using classes but single files for each "functional group".

I will add an example of the file structure I have setup, but this should not mean that my proposal should be implementes. It is mean as a starting point to discuss how to split mobilflight.cpp and to habe an issue to keep track on this topic.
image
If one would like to have a deeper look into the file structure and the files itself, see my branch https://github.com/elral/MobiFlight-FirmwareSource/tree/multiple_files

elral:
Fixes #148

1. mobiflight.cpp is splitted into several files, one per device plus commandmesenger.cpp and config.cpp.
2. All device wrapper classes (e.g. Button.cpp) are initialized dynamically using the new deviveBuffer and there corrospending classes (e.g. MFBUtton.cpp) are intialized statically (but they will still initialized in the deviceBuffer). This gives the lowest RAM and flash usage. One exception is the MFStepper.cpp class, this class is still intialized dynamically to differ between FULL4WIRE and DRIVER support.
3. The functionality of MFEEPROM.init() was moved to the constructor.
4. #ifdef DEBUG was renamed to #ifdef DEBUG2CMDMESSENGER to avoid conflicts with other libraries (not in the used one, but maybe later in other libraries)
5. The loop() is restructured, checking the time if a device must be updated is now done within the loop() and not the functions which are called before. I see the loop() as a kind of "controller".
6. in all include files #ifndef xyz / #define xyz was replaced by #pragma once
7. All files were checked if the include files are required. If they are not required they got deleted
8. I hope I get it all

As I do not own Input- and Outputshifters I would highly appreciate if somebody could test it. All other devices are working. Also most changes from the last merging must be implemeted by hand.
I am still wondering about the uncommented PowerSaveOutputs(state); in the actual branch (what I left for now). Do outputs does not have a power saving?


neilenns:
@elral This has a build failure:

```
src/CommandMessenger.cpp:3:10: fatal error: commandmessenger.h: No such file or directory
```

elral:
> @elral This has a build failure:
> 
> ```
> src/CommandMessenger.cpp:3:10: fatal error: commandmessenger.h: No such file or directory
> ```

Holy moly, (my?) PlatformIO does not consider lowercase / capitel letter. Thanks for your hint! I will change the file accordingly, otherwise I had to modify all files where it is included.

neilenns:
@elral Windows vs. Linux build environments I think.

Use the `git mv` command to rename it. See https://makolyte.com/how-to-change-the-casing-of-a-filename-when-using-git-on-windows/

elral:
@neilenns do you have also a hint for me how to rename a folder? I didn't see a capitel "I" in `./src/MF_LCDDIsplay`.
All time new surprises ;)

elral:
@neilenns thanks a lot! I will care about tomorrow.

elral:
I will create a new pull request after I opened issues for the changes which were also implemented (as per Neils recommendation)

elral:
@MobiFlight-Admin @neilenns @GioCC 
I checked again my pull request and plan to re-submit this one after discussions with Sebastian.
While going through all the files again (e.g. checking upper/lower case), there is still a topic I do not understand. Maybe someone can guide me on this.

Within the actual branch there is a function each for both Shifters:
* OnInitInputShifter()
* OnInitOutputShifter()

But as far as I can see both functions are never used, is this correct? Or do I simply not find the file where they are used?
Any help would be appreciated.

GioCC:
From what I can see, `OnInitOutputShifter()` was meant to be associated with a command, which however has never been defined; it can either be defined now, or the function can safely be removed.
I can't find `OnInitInputShifter()` though, and it would make little sense anyway (inputs need not be initialized, possibly read the first time, but that's done anyway).

elral:
> I can't find `OnInitInputShifter()` though, and it would make little sense anyway (inputs need not be initialized, possibly read the first time, but that's done anyway).

Ups, it's better to use copy and paste and to type it manually. It must be: `OnInitInputShiftRegister`

DocMoebiuz (approved):
BIG THANK YOU TO EVERYONE!!!

PR #156, opened 2022-02-22
Fix broken output shifters

### LINKED ISSUE #155: Output shifters are broken with beta firmware
Offending line is here: https://github.com/MobiFlight/MobiFlight-FirmwareSource/blob/46969cae100fd93189a51e443c935926a9234f80/src/MF_Modules/MFOutputShifter.cpp#L91

It's off by one, it should be using `i - 1` for the index into the `outputBuffer` array.

neilenns:
Fixes #155

## Description of changes

Adds a `- 1` to the index into `_outputBuffer` so the correct values get sent to the shift register chips.

### DIFF
--- a/src/MF_Modules/MFOutputShifter.cpp
+++ b/src/MF_Modules/MFOutputShifter.cpp
@@ -16,7 +16,6 @@ void MFOutputShifter::setPin(uint8_t pin, uint8_t value, uint8_t refresh)
 
   uint8_t idx = (pin & 0xF8) >> 3;
   uint8_t msk = (0x01 << (pin & 0x07));
-
   if (value > 0)
   {
     _outputBuffer[idx] |= msk;
--- a/src/MF_Modules/MFOutputShifter.cpp
+++ b/src/MF_Modules/MFOutputShifter.cpp
@@ -11,25 +11,32 @@ MFOutputShifter::MFOutputShifter()
 
 void MFOutputShifter::setPin(uint8_t pin, uint8_t value, uint8_t refresh)
 {
-  if (!_initialized) return;
+  if (!_initialized)
+    return;
 
-  uint8_t idx = (pin & 0xF8)>>3;
+  uint8_t idx = (pin & 0xF8) >> 3;
   uint8_t msk = (0x01 << (pin & 0x07));
-  //if(idx < _moduleCount) return;
-  if (value > 0) {
-    _outputBuffer[idx] |= msk;  
-  } else {
-    _outputBuffer[idx] &= ~msk;  
-  }    
-  if(refresh) updateShiftRegister();
+
+  if (value > 0)
+  {
+    _outputBuffer[idx] |= msk;
+  }
+  else
+  {
+    _outputBuffer[idx] &= ~msk;
+  }
+  if (refresh)
+    updateShiftRegister();
 }
 
-void MFOutputShifter::setPins(char* pins, uint8_t value)
+void MFOutputShifter::setPins(char *pins, uint8_t value)
 {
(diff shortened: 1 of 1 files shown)


src/MF_Modules/MFOutputShifter.cpp
```diff
@@ -46,49 +53,52 @@ void MFOutputShifter::attach(uint8_t latchPin, uint8_t clockPin, uint8_t dataPin
...
 {
-   digitalWrite(_latchPin, LOW);
-   for (uint8_t i = _moduleCount; i>0 ; i--) {
-      shiftOut(_dataPin, _clockPin, MSBFIRST, _outputBuffer[i]); //LSBFIRST, MSBFIRST,
-   }    
-   digitalWrite(_latchPin, HIGH);
+  digitalWrite(_latchPin, LOW);
+  for (uint8_t i = _moduleCount; i > 0; i--)
+  {
+    shiftOut(_dataPin, _clockPin, MSBFIRST, _outputBuffer[i - 1]); // LSBFIRST, MSBFIRST,
```
neilenns, line 101: This is the actual change. All the rest is whitespace formatting done by VSCode to clean it up.

GioCC (approved):

src/MF_Modules/MFOutputShifter.cpp
DocMoebiuz: either we share all the same formatting rules or you should not reformat everything if there is no need.

your formatting is not necessarily my preference, e.g. i like the curly braces in the same line where the if is. But most importantly it should be aligned with everyone.

neilenns: I'm not doing anything intentionally, it's just VSCode doing it. I wonder if I have a formatter installed? I'll check.

neilenns: I think it's probably part of the VSCode C++ extensions? I don't see any explicit formatter, but it's definitely changing things on save. I'm trying to update the PR with just the code change, hang tight.

GioCC: In order to avoid frequent style discrepancies, we could conveniently use the automatic formatter of the "C/Cpp" plugin of VSCode (in `settings.json`).
If you want to give it a try, my settings (slightly modified from the defaults) are:
`"C_Cpp.clang_format_fallbackStyle": "{ BasedOnStyle: LLVM, UseTab: Never, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Linux, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, NamespaceIndentation: All, FixNamespaceComments: false, AlignConsecutiveAssignments: AcrossComments, AlignConsecutiveBitFields: AcrossEmptyLinesAndComments, AlignConsecutiveDeclarations: AcrossEmptyLinesAndComments, AlignConsecutiveMacros: AcrossComments, AlignTrailingComments: true, AllowShortBlocksOnASingleLine: Empty, AllowShortEnumsOnASingleLine: true, AllowShortFunctionsOnASingleLine: Inline, AllowShortIfStatementsOnASingleLine: WithoutElse,  BreakInheritanceList: BeforeColon, EmptyLineBeforeAccessModifier: LogicalBlock, IndentPPDirectives: BeforeHash }"
`

neilenns: I fully support an automatic formatter that everyone has access to via VSCode. I don't care at all what the settings are, I just want to know when I hit "save" things will format to however is the right way to be consistent with everyone else.

PR #160, opened 2022-02-25
avoid crash if string is longer then columns, Fixes #158

### LINKED ISSUE #158: Configuration of a LCD with 8x2 characters leads to crash of the Arduino
After configuration of a LCD with 8x2 characters the firmware crashes on bootup.
Within the test() function from MFLCDDisplay.cpp more characters then fit on the LCD are written.

elral:
Check if string to be printed is longer than available columns. If so an "overflow" of 'c' happens and a lot of spaces are printed to the LCD which seems to crash the Arduino. By checking this condition and not printing spaces this crash is avoided.

I thought some time to check this in the LiquidCrystal_I2C library, but that would mean a lot of modifications as the column is not tracked there. The test() function here is the only place where this could happen so it's easier to make the modifications here.

Fixes #158

### DIFF
--- a/src/MF_Modules/MFLCDDisplay.cpp
+++ b/src/MF_Modules/MFLCDDisplay.cpp
@@ -81,20 +81,19 @@ void MFLCDDisplay::test()
 
 void MFLCDDisplay::_printCentered(const char *str, uint8_t line)
 {
-  _lcdDisplay->setCursor(0, line);
+  uint8_t startCol = 0;
+  uint8_t printChar = _cols;
+  
   if (_cols > strlen(str))
   {
-    for (byte c = 0; c != ((_cols - strlen(str)) / 2); c++)
-    {
-      _lcdDisplay->print(F(" "));
-    }
-    _lcdDisplay->print(str);
-    for (byte c = 0; c != ((_cols - strlen(str)) / 2); c++)
-    {
-      _lcdDisplay->print(F(" "));
-    }
-  } else
+    startCol = (_cols - strlen(str)) / 2;
+    printChar = strlen(str);
+  }
+
+  _lcdDisplay->setCursor(startCol, line);
+  
+  for (uint8_t i = 0; i < printChar; i++)
   {
-    _lcdDisplay->print(str);
+    _lcdDisplay->write(str[i]);
   }
 }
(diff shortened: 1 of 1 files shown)


GioCC:
I cannot verify it right now, but a few questions:
- if `(_cols - strlen(str))` is an odd number, one space less than necessary is printed;
- it would definitely be advisable to use "<" rather that "!=" in the for condition;
- doesn't the `_lcdDisplay->print(str)` cause any issue if `str` is longer than `_cols `anyway? Otherwise the only difference from the previous situation could be that the underflow in `(_cols - strlen(str))` caused a much longer sequence of spaces to be printed, but still in excess.

I might suggest a safe `strncpy()` in a local buffer to make sure that the printed string is positively not longer than allowed.

GioCC (commented):
I cannot verify it right now, but a few questions:
- if `(_cols - strlen(str))` is an odd number, one space less than necessary is printed;
- it would definitely be advisable to use "<" rather that "!=" in the for condition;
- doesn't the `_lcdDisplay->print(str)` cause any issue if `str` is longer than `_cols `anyway? Otherwise the only difference from the previous situation could be that the underflow in `(_cols - strlen(str))` caused a much longer sequence of spaces to be printed, but still in excess.

I might suggest a safe `strncpy()` in a local buffer to make sure that the printed string is positively not longer than allowed.

GioCC:
On second thought, a local buffer, which should be sized on the max possible line length, could cause weird RAM issues if the stack is not large enough; on the other hand, printing char by char might be _very_ slow (I should have a look of what happens inside `_lcdDisplay->print`). Probably the solution needs further investigation.

elral:
@GioCC I used already a "<" condition in the if statement, does a logical expression show up for you? If the string is longer as the columns, then do not calculate the number of spaces was my intention.

The `->print(str)` leads to `write(str)` and this to `write(char)`, so in the end printing a single character in the test function ends up in the same printing the string.

Sending "some" more characters then the display has columns does still work (see feedback from discord), but maybe it would be really better to write single characters.

GioCC:
> @GioCC I used already a "<" condition in the if statement, does a logical expression show up for you?

No, I mean in the "for" statement - it wasn't you, it was that way already in the existing code.

> Sending "some" more characters then the display has columns does still work (see feedback from discord), but maybe it would be really better to write single characters.

Exactly, probably the issue arises if you print _way too many_ characters (e.g. : _cols=8, str len =16, diff = 0xF8 -> ends up printing 248 chars).

> /edit: just saw that I mixed up the order of `_cols` and `strlen(str)`, or that I have to use a `>=`

You mean line 85? I think that's correct

elral:
We don't even need the `for` loops to print the spaces before and! after printing the string. Just calculating the start postion is OK as the LCD is cleared before calling this function. And the printing single characters not exceeding the number of columns.

DocMoebiuz (approved):
Thanks!

PR #167, opened 2022-03-12
Code reformatting and reorganization

GioCC:
Note: This PR does not include any change in features or in code other than reformatting and file / source reorganization (which are partly meaningful in itself and partly used in several subsequent PR to be issued soon).
It has been submitted separately in order to help tell purely formal changes apart from feature changes (which will be in other PRs). 
Since the changes are exclusively formal, there is no associated issue.

## Description of changes
Rearrangement of files:
- "Include" directory removed (.h files moved to "src")
- MFEEPROM.h moved to MF_Modules (.cpp was under "_Boards/Atmel")
Other:
- Device type constants moved to "config.h"
- Header inclusion optimized (main includes collected in "mobifligt.h)
- Code formatting and headers made uniform
- introduced VS Code formatting config (settings.json)


GioCC:
Looks like automatic build fails (local build is fine) because of this:
```
Library Manager: Warning! Could not install dependency {'name': 'Adafruit-MAX31855', 'frameworks': ['arduino']} 
for package 'CmdMessenger'
```
causing: 
```
In file included from src/mobiflight.h:13:0, from src/CommandMessenger.cpp:7:
src/commandmessenger.h:9:10: fatal error: cmdmessenger.h: No such file or directory
```

Anybody knows how to fix this?

elral:
> * MFEEPROM.h moved to MF_Modules (.cpp was under "_Boards/Atmel")
MFEEPROM.cpp was put intentionally under _Boards/Atmel, as this file if the only one which is hardware dependent. For other hardware there would be a new folder like e.g. _Boards/RaspiPico where the relevant MFEEPROM.cpp file for this hardware would be stored. And in the platformio.ini there is the hardware dependent include path defined.

Regarding the built error, could it be something with upper/lower case? Builds on github are case sensitive what I have learned during my last pull request...

neilenns:
That's my guess as well, the case needs to match.


GioCC:
> > * MFEEPROM.h moved to MF_Modules (.cpp was under "_Boards/Atmel")
> 
> MFEEPROM.cpp was put intentionally under _Boards/Atmel, as this file if the only one which is hardware dependent. For other hardware there would be a new folder like e.g. _Boards/RaspiPico where the relevant MFEEPROM.cpp file for this hardware would be stored. And in the platformio.ini there is the hardware dependent include path defined.
> 

I understand that, but I don't think it's a particularly correct choice, for a number of reasons.

For starters, MF is currently 99% directed to AVR Arduino boards, and that doesn't seem very likely to change much anytime soon.
Even if "official" versions for other types of hardware do exist (basically @neilenns's one for the Pico, which is the only one we currently have at hand), others are/will be experimental projects, which will most likely require significant rework anyway. 

Generally speaking, and regardless of their diffusion, it's not guaranteed that for other platforms the required changes can be included simply by isolating a few non-core files to platform-specific directories.

Platform changes often require _#ifdefs_ on short sections of code within the core anyway (BTW this will be the case for the modifications I have not yet posted regarding the elimination of index arrays). 
In this case, having several - probably very similar - copies of a same file is impractical and possibly even an insidious source for bugs.
Besides, these _#ifdef_'s would not even require changing the build scripts (platformio.ini in our case), because they are predefined for the specific platform (e.g. "\_\_AVR\_\_" for gcc).

Another specific reason is because the MFEEprom class is based on the Arduino EEPROM class, which is a framework class: for instance, the Pico version of MFEEPROM  is not really platform-specific, it just uses a version of EEPROM that is VERY slightly different from the standard one.
A more correct approach would be for the implementation to supply an interface-compatible EEPROM class, and even where there are small differences, these can be easily resolved with local _#ifdef_'s. BTW, in the case of the Pico version, this is not even required (versions are almost identical), a pair of short _#ifdef_ s would be all it takes.

For the above reasons, having to deal with individual files tucked away in platform-specific subdirectories seems to me an unnecessary nuisance.


> Regarding the built error, could it be something with upper/lower case? Builds on github are case sensitive what I have learned during my last pull request...

Right, that was it! Didn't think of that, thanks for the tip!

_Boards/Atmel/Board_Mega/MFBoards.h
```diff
@@ -24,29 +30,31 @@
...
-#define EEPROM_SIZE             4096    // EEPROMSizeMega
-#define MEMLEN_CONFIG           1496    // max. size for config which wil be stored in EEPROM
-#define MEMLEN_CONFIG_BUFFER    1000    // max. size for configBuffer, contains only names from inputs
-#define MF_MAX_DEVICEMEM        1500    // max. memory size for devices
+#define MOBIFLIGHT_TYPE     "MobiFlight Mega"
+#define MOBIFLIGHT_SERIAL   "1234567890"
+#define MOBIFLIGHT_NAME     "MobiFlight Mega"
+#define EEPROM_SIZE         4096 // EEPROMSizeMega
+#define MEMLEN_CONFIG       1496 // max. size for config which wil be stored in EEPROM
+#define MEMLEN_NAMES_BUFFER 1000 // max. size for configBuffer, contains only names from inputs
```
DocMoebiuz, line 53: Why do we have this change as part of the reformatting

DocMoebiuz (approved):
Wow.... a lot of reformatting. thanks!

DocMoebiuz:
@GioCC why did you decide to take 4 spaces as indentation, is it because of C#? We could have saved quite some changes. 
I also noticed that some methods() are now methods(void), others are not. Is there a specific logic behind that?

GioCC:
Maybe too late to reply, but:
- I have fixed spacing of 4 set in all my editors, so it has appaerently taken this up also in reformatting
- As a rule, I tend to add the (void) in declaration and definitions, for more correctness (even if that's optional) but I also find it practical to quickly identify decl/def from calls in searches

PR #171, opened 2022-04-22
Config gets not "resetted" if InputShifter is deactivated

### LINKED ISSUE #170: Config gets not "resetted" if Input Shifter support is deactivated
In config.cpp `configLength `and `configActivated `will be resetted only if `MF_INPUT_SHIFTER_SUPPORT == 1`

In `void resetConfig()` `configLength = 0;` and `configActivated = false;` must not be encapsulated with `MF_INPUT_SHIFTER_SUPPORT == 1`

elral:
Fixes #170

configLenght=0 and configActivated=false must not be encapsulated within MF_INPUT_SHIFTER_SUPPORTT==1

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -146,9 +146,9 @@ void resetConfig()
 #endif
 #if MF_INPUT_SHIFTER_SUPPORT == 1
     InputShifter::Clear();
+#endif
     configLength    = 0;
     configActivated = false;
-#endif
 }
 
 void OnResetConfig()
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -147,6 +147,9 @@ void resetConfig()
 #if MF_INPUT_SHIFTER_SUPPORT == 1
     InputShifter::Clear();
 #endif
+#if MF_DIGIN_MUX_SUPPORT == 1
+    DigInMux::Clear();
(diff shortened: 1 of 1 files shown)


elral:
@GioCC is it intended not to clear the multiplexer within `resetCinfig()`?

GioCC:
> @GioCC is it intended not to clear the multiplexer within `resetConfig()`?

You're right (in spotting the bug), it's NOT meant that way. I had noticed this issue too, but I mistakenly thought it was a merging glitch (which it probably was) that had not made its way into the final PR. Looks like I should have checked better!

elral:
So should I close this pull request and you issue a new one with the above changes and required additions (clear multiplexers)?

GioCC:
> So should I close this pull request and you issue a new one with the above changes and required additions (clear multiplexers)?

It would probably make sense if you go ahead and add it yourself... I see no point in closing and reopening an issue for the same purpose.
Or if you prefer to keep issues separate, you can correct this one and then I open a different issue to add the Mux clear, but that would probably be unnecessarily complicated.

GioCC:
Thanks @elral!

DocMoebiuz (approved):
Thanks!

Issue #172, opened 2022-04-23
Improvement of EEPROM/Config handling


GioCC:
Slight improvement of code for EEPROM and config handling:
- removes burden of unnecessary parameter passing
- collects config parameters to a single struct for better encapsulation 
- improves readability a little

DocMoebiuz:
@GioCC please make sure to include `fixes #issue-number` in your comment in the PR

PR #173, opened 2022-04-23
Improvement of pointer handling in EEPROM functions for config parsing

GioCC:
## Description of changes

Addresses #172:
`config` parameters collected in a single struct for better encapsulation and readability
Added EEPROM function aliases to avoid unnecessary parameter passing (and improve readability)
Some variable names changed for consistency


DocMoebiuz (approved):
Thanks!

Issue #174, opened 2022-04-23
Removal of the fixed per-device limit of configurable devices (and associated data structures)


GioCC:
This issue refers to discussion #154.
Basically, the "index" arrays for each device type are suppressed, and devices are allocated completely in the already existing object  buffer.
This way, there is no longer a limit to the number of devices of a given type that can be allocated (e.g.: one can allocate exclusively shift registers, as many as the avaliable space allows). The max amounts need no longer to be defined.

DocMoebiuz:
@elral can you please also comment on this? I think it would be great to have this.

Open questions:
* is actually no limitation for devices always sensible? Do we not also depend on refresh speed? I think it would be OK to not worry about this and leave it up to the user to decide what is acceptable.
* Does the UI need any modifications? Right now it knows that it cannot add an additional device. This would have to be modified because we don't upload the config immedieately every time we add a device. So the UI would have to keep track of the used RAM too.

elral:
This would give complete flexibility to the user which and how much! devices he want to use on a board.
As every device is affected this will be a lot of changes and we should do tests before merging.
But I am also thinking about if processing time could be affected if e.g. a user defines on a Mega "a couple" of inputshifter/multiplexer. Do we want to allow a mega with input shifters on all pins? Or more than 2 LCD's as they need much more time than we need for reading the encoders. So it might be possible that encoders are not read proberbly if lot's of LCD are connected (limited by the available I2C addresses).

I will not say that it should not be done, but I think that still a limitation of max. devices is required.

On the connector side I these the modifications as Sebastian has already mentioned. For now the firmware reports a `kStatus` message that the buffer is full, but I think this is not evaluated for now in the connector. Maybe this should be changed in a new `kError` message or at least processing the existing buffer full message.

GioCC:
I fully second all of @elral 's comments. Even though we of course don't want to allow a mega with input shifters on all pins, and however unlikely we assume that to be, this would be definitely a possibility unless restrained.
"Heavy" peripherals like LCDs could probably be limited to a reasonable number anyway (whether that means keeping the current storage implementation or handling them the new way), but more "common" peripherals may add up to unwanted scan times.
I also find @DocMoebiuz 's proposal basically correct ("leave it up to the user to decide what is acceptable"), maybe with some safeguard limit added.

[...]

PR #175, opened 2022-04-23
Modifications to get rid of device index arrays

GioCC:
### Description of changes
_Addresses issue #174 (which in turn refers to discussion #154)._
The "index" arrays for each device type are suppressed, and devices are allocated completely in the already existing object  buffer; this way, there is no longer a limit to the number of devices of a given type that can be allocated (e.g.: one can allocate exclusively shift registers, as many as the avaliable space allows). The max amounts need no longer to be defined.

### Notes
- Modifications are broad (also because they involve all devices), but mostly fairly repetitive.
Also, the refactoring contributes IMHO to a certain "modularization" of the structure of device-related code.
- The current `allocateMem` class is replaced by a slightly more complex `stowManager`, which has the same basic purpose but with more extended functions.
- There is a (slight) gain in RAM, although this is not the actual goal.

___NOTE: This is a draft PR; the code, albeit complete, still requires testing and cleanup (and obviously also review).___
If interested, please don't forget to have a look at the comments  in the `IODevice` class, which is pivotal.

I will greatly appreciate any checking effort and suggestions.


elral:
Additional to my comment above, changes in config.cpp are required. Can't do it in the code, as this file is not changed within this Pull Request. Revert Back from pullrequest #173 is not merged here:

```
bool readRecordTailFromEEPROM(char **dest = NULL, char *cap = (char *)0xFFFF)
{
    char temp = 0;
    do {
        temp = MFeeprom.read_char(); // read the first character
        if (dest) {
            *((*dest)++) = temp;
            // TODO: this function should be buffer-agnostic -
            //  the specific constant for buffer size should not be hardcoded!
            if (*dest >= cap) break; // nameBuffer full: stop copying
        }
    } while (temp != ':'); // reads until limiter ':' and locates the next free buffer position
    if (dest) {
-------------------------------
//        *((*dest)--) = 0x00; // replace ':' by NULL, terminates the string
        *((*dest)-1) = 0x00; // replace ':' by NULL, terminates the string
    }
//    return (*dest >= cap);
    return (*dest <= cap);
------------------------------
}
```
I am not completly sure about `*((*dest)-1) = 0x00; // replace ':' by NULL, terminates the string`

Nevertheless my mega2560 crashes even with this changes. W/o them the config gets not loaded and the mega2560 doesn't crash.

elral:
Hmhm, tested it again. Now the mega does not crash!?! Please don't ask me what I did wrong before. Sorry.
But it seems that the config get's still not loaded. No input and output device is working...

GioCC:
> Hmhm, tested it again. Now the mega does not crash!?! Please don't ask me what I did wrong before. Sorry. But it seems that the config get's still not loaded. No input and output device is working...

Ok, I think I'll have to take a step back and review it again myself first. I have been trying to follow late changes (basically PRs from #171 on) and it's getting a little confusing; I'd like to have a look to last-minute changes because there's definitely something that doesn't add up. Thanks for your effort so far, I will get back to you all once I figure out where it's at.

elral:
What I found so far, `Button::Add(params[0], nameBufPtr);` leads to a crash.
With the above mentioned changes the `readConfig()` will be finished (w/o button configured), but then it seems to crash.

PR #182, opened 2022-05-10
Device Buffer get's not cleared if a config is uploaded

### LINKED ISSUE #181: Device Buffer get's not cleared if a config is uploaded
If a config gets changed and uploaded, the existing config will be resetted, the changed config gets uploaded and at leat this "new" config gets loaded.
Within resetting the existing config, the device buffer gets not cleared and with loading the "new" config some/all devices will not fit into the remaining space of the device buffer and therefore not be loaded.

elral:
After changing and uploading a new config the devive buffer gets not cleared. Added the missing function solves this issue.

Fixes #181

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -152,6 +152,7 @@ void resetConfig()
 #endif
     configLength    = 0;
     configActivated = false;
+    ClearMemory();
 }
 
 void OnResetConfig()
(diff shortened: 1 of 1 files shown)


elral:
Please have a special look on the comments to the function `void OnSaveConfig()`. For me it was helpfull to test things about config "upload" w/o connected to the GUI. Debug prints from the firmware are not displayed in the terminal from the GUI, therefore I have to test it with a serial terminal.

If the comments are not desired, I will delete them and update the pull request.

DocMoebiuz (approved):
Thanks!

PR #183, opened 2022-05-10
Increase number of encoders for Uno and ProMicro

### LINKED ISSUE #180: Increase number of encoders for Uno and ProMicro
As proposed in https://github.com/MobiFlight/MobiFlight-Connector/issues/784 the max. number of encoders could be increased to 9 to use all available pins from Uno ans ProMicro. This will require additional 8 Bytes which is OK due to all memory optimazations which were implemented in the past.

elral:
The max. number of encoders is increased to 9 for Uno and ProMicro. All available 18 Pins can be used for encoders.

Fixes #180


DocMoebiuz (approved):
Great. Thanks!

PR #185, opened 2022-05-22
platformio.ini uses deprecated `src_filter`

### LINKED ISSUE #184: platformio.ini uses deprecated `src_filter`
In the meantime `src_filter` which is used in the platformio.ini file is deprecated and gives a warning when firmware is compiled.

Replace `src_filter` by `build_src_filter` to avoid this warning.

elral:
To avoid warning of deprecated function replace src_filter by build_src_filter

Fixes #184


DocMoebiuz (approved):
Awesome! Thank You!

PR #187, opened 2022-06-08
Optimization of MFEEPROM class

### LINKED ISSUE #186: Optimization of MFEEPROM class
@GioCC did already some improvements to the MFEEPROM class, but they were reverted. While playing around with a Raspberry Pico I found additional topics:

1. If writing/reading a block of data to/from the eeprom while adressing the last memory location, the complete block gets not written/read. 
2. `MFEEPROM` is defined as a global class which depends on another global class `EEPROM`. It can not be guaranteed that the order of both calls are always in the correct order. For the Arduino Framework it doesn't seem to have any effect (but can not be ensured that this may never happen in the future). Using other hardware with there frameworks can lead to not working code. And it's not a legal C++ code.

elral:
1. Like @GioCC always proposed, read/write block and write byte returns true or false (not used for now)
2. private variable `eepromLength ` is cahnged to `_eepromLength ` like in other MF classes.
3. If last eeprom location was addressed at read/write block complete block was not written/read
4. `MFEEPROM `needs separate `init()` for C++ compliance. `MFEEPROM `and `EEPROM `are global classes, and "correct" order of constructors can not be ensured. This is more a relevant for additional hardware/frameworks, but it might also a topic for the  future.

I am also fine if @GioCC includes it in his PR #175 (some are already included).

Fixes #186

### DIFF
--- a/src/MF_Modules/MFEEPROM.cpp
+++ b/src/MF_Modules/MFEEPROM.cpp
@@ -22,8 +22,8 @@ uint16_t MFEEPROM::get_length(void)
 
 bool MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
 {
+    if (adr + len > _eepromLength) return false;
     for (uint16_t i = 0; i < len; i++) {
-        if (adr + len >= _eepromLength) return false;
         data[i] = read_char(adr + i);
     }
     return true;
--- a/src/MF_Modules/MFEEPROM.cpp
+++ b/src/MF_Modules/MFEEPROM.cpp
@@ -31,8 +31,8 @@ bool MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
 
 bool MFEEPROM::write_block(uint16_t adr, char data[], uint16_t len)
 {
+    if (adr + len > _eepromLength) return false;
     for (uint16_t i = 0; i < len; i++) {
-        if (adr + i >= _eepromLength) return false;
         EEPROM.put(adr + i, data[i]);
     }
     return true;
--- a/src/MF_Modules/MFEEPROM.cpp
+++ b/src/MF_Modules/MFEEPROM.cpp
@@ -8,41 +8,47 @@
 #include "MFEEPROM.h"
 #include <EEPROM.h>
 
-MFEEPROM::MFEEPROM()
+MFEEPROM::MFEEPROM() {}
+
+void MFEEPROM::init(void)
 {
-    eepromLength = EEPROM.length();
+    _eepromLength = EEPROM.length();
 }
 
 uint16_t MFEEPROM::get_length(void)
 {
-    return eepromLength;
+    return _eepromLength;
 }
 
-void MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
+bool MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
 {
     for (uint16_t i = 0; i < len; i++) {
+        if (adr + len >= _eepromLength) return false;
         data[i] = read_char(adr + i);
     }
+    return true;
 }
 
-void MFEEPROM::write_block(uint16_t adr, char data[], uint16_t len)
+bool MFEEPROM::write_block(uint16_t adr, char data[], uint16_t len)
 {
-    if (adr + len >= eepromLength) return;
     for (uint16_t i = 0; i < len; i++) {
+        if (adr + i >= _eepromLength) return false;
         EEPROM.put(adr + i, data[i]);
     }
+    return true;
 }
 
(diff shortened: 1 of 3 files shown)


GioCC:
About .3: nice catch. I was wondering whether, instead of aborting R/W completely, it could have made sense to read or write just the "legal" part of the range; however, for data consistency, I believe you made the right choice.

About .4:  Also, sharp observations. In the specific case, being EEPROMclass (and EEptr) stateless and constructorless structs (and static at that), there should be no initialization issue; however it's still a correct way of thinking. 
Sometimes the late initialization is flagged as a discouraged practice (there's a paradigm, I miss the exact definition right now, that states that objects should be initialized upon creation and before any access is possible), however embedded applications have their singularities, and they are actually one of the main exception to this principle.
Thanks!

src/MF_Modules/MFEEPROM.cpp
```diff
@@ -8,41 +8,47 @@
...
 {
-    return eepromLength;
+    return _eepromLength;
 }
 
-void MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
+bool MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
 {
     for (uint16_t i = 0; i < len; i++) {
+        if (adr + len >= _eepromLength) return false;
```
GioCC, line 26: This line could be moved before the for loop

elral:
Yes, that is what I mean with the changes you already made in the Pull Request, but this one is reverted. I am also fine if you include it in your pulrequest #175 or if you reopen your pullrequest.

But `if (adr + len >= EEPROM.length()) return false;` within `write_block()` is still wrong, it must be without `=`.
On a 1kB eeprom, starting on adress 0, 1024 Bytes can be written. That means `(0 + 1024) >= 1024` aborts writing to the eeprom.
That is my main intention beside having the init() function (like my raspberry pico branch showed me).

Having the if clause within the for loop (everything gets written until end of eeprom) or before (abort the complete block if the last eeprom location is adressed) is something which has to be agreed. I think it's more a philosophy question.

GioCC (approved):

src/MF_Modules/MFEEPROM.cpp
```diff
@@ -8,41 +8,47 @@
...
 uint16_t MFEEPROM::get_length(void)
 {
-    return eepromLength;
+    return _eepromLength;
 }
 
-void MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
+bool MFEEPROM::read_block(uint16_t adr, char data[], uint16_t len)
 {
+    if (adr + len > _eepromLength) return false;
```
DocMoebiuz, line 25: is the check > correct, in other locations we are using >=

I think it should be >=

src/MF_Modules/MFEEPROM.cpp
```diff
@@ -8,41 +8,47 @@
...
         data[i] = read_char(adr + i);
     }
+    return true;
 }
 
-void MFEEPROM::write_block(uint16_t adr, char data[], uint16_t len)
+bool MFEEPROM::write_block(uint16_t adr, char data[], uint16_t len)
 {
-    if (adr + len >= eepromLength) return;
+    if (adr + len > _eepromLength) return false;
```
DocMoebiuz, line 34: Same as above

src/MF_Modules/MFEEPROM.cpp
```diff
@@ -8,41 +8,47 @@
...
     for (uint16_t i = 0; i < len; i++) {
         EEPROM.put(adr + i, data[i]);
     }
+    return true;
 }
 
 char MFEEPROM::read_char(uint16_t adr)
 {
-    if (adr >= eepromLength) return 0;
+    if (adr >= _eepromLength) return 0;
```
DocMoebiuz, line 43: here we are using >=

DocMoebiuz (changes requested):
Looking good, just one observation, please review.

src/MF_Modules/MFEEPROM.cpp
GioCC: That was addressed in [@elral's comment above](https://github.com/MobiFlight/MobiFlight-FirmwareSource/pull/187#issuecomment-1149908734), he was right in spotting the >/>= issue.

src/MF_Modules/MFEEPROM.cpp
GioCC: This is different, `adr` only should be 0...`len-1`.

GioCC:
BTW, please consider reviving #173.... it was quickly merged but hastily removed because of a small bug (which caused an evident malfunction), but that bug has long since been corrected.

DocMoebiuz (approved):

PR #189, opened 2022-06-10
Add firmware debug printouts to log in the UI

### LINKED ISSUE #188: Add firmware debug printouts to log in the UI
Firmware printouts during debuging should show up in the log/terminal from the UI.
Until now they are marked as `kStatus` and they will not show.
A separate identifier should be choosen to mark them as debug info.

The connector must be changed accordingly, see issue #854 in the Connector repo

elral:
Debug prints send via the commandmessenger are marked with kStatus. This is changed to KDebug to seperate them from status messages.
With changes in the UI the will appear if loglevel is debug.

Fixes #188

### DIFF
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -80,7 +80,7 @@ void attachCommandCallbacks()
 #endif
 
 #ifdef DEBUG2CMDMESSENGER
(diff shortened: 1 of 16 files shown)


DocMoebiuz (approved):
Thanks! You're awesome!

PR #191, opened 2022-07-22
Improvements to config management functions

### LINKED ISSUE #190: Improvements to config management functions
At the start of the configuration upload process, the `ResetConfig` command from the connector immediately clears the device object variables (arrays and bulk storage area) and also "clears" the configuration stored in EEPROM (by zeroing the first byte) in the same place. The two operations, however, are completed in other functions at later moments; the call structure, while working as it is (relying on the correct call sequence from the connector), could bring to inconsistencies in data.

The `_activateConfig()` function can be integrated in `ReadConfig()` (they are always called together) to further improve consistency.

Function naming is also unclear in some place and could use a little improvement.

GioCC:
## Description of changes

Fixes #190.

The PR makes following changes:
- separate device object data clearing from EEPROM data clearring into different functions
- have the object clearing function called at the correct time, when the EEPROM config is read back, making the operation consistent
- remove `_activateConfig()` (_always_ called in pair after `ReadConfig()`) and move config activation flag setting at the end of `ReadConfig()`
- rename some functions with more significant names

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -63,9 +63,9 @@ char       nameBuffer[MEM_LEN_CONFIG]    = "";
 uint16_t   configLength                  = 0;
 boolean    configActivated               = false;
 
-void resetConfig();
-void readConfig();
-void _activateConfig();
+void resetDeviceConfig();
+void resetStoredConfig();
+void readStoredConfig();
 
 // ************************************************************
 // configBuffer handling
@@ -96,13 +96,13 @@ void loadConfig()
     cmdMessenger.sendCmd(kStatus, F("Load config"));
 #endif
     if (readConfigLength()) {
-        readConfig();
-        _activateConfig();
+        readStoredConfig();
     }
 }
 
 void OnSetConfig()
 {
+    // Append a chunk of configuration text to EEPROM storage
 #ifdef DEBUG2CMDMESSENGER
     cmdMessenger.sendCmd(kStatus, F("Setting config start"));
 #endif
@@ -121,8 +121,12 @@ void OnSetConfig()
 #endif
 }
 
-void resetConfig()
+void resetDeviceConfig()
 {
+    // Clear device variables
+
+    configActivated = false;
+
     Button::Clear();
     Encoder::Clear();
     Output::Clear();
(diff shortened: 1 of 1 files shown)


GioCC:
Ok, as requested, these a few explanations regarding this PR.

`resetConfig()`->`resetDeviceConfig()` + `resetStoredConfig()`:
allows to separate the function that only clears the internal variables (internal housekeeping operation that occcurs normally in a configured board, leaving the config unaltered) from the one that makes a "clean slate" of the board.  
Besides being IMHO conceptually significant, the true purpose is to allow inserting only the `resetDeviceConfig()` part into `readStoredConfig()`.  
This latter function assumes that the data to be read from the EEPROM is used to build a variable set from scratch; this precondition however is not enforced (it is implicitly, because it "just happens" that the variables are cleaned elsewhere).
This might lead to a sneaky issue if the code is changed (even inadvertently), and the "new" devices could be added to the existing ones.

`_activateConfig()`: this function (even as a command) does no longer make much sense, since the "activation" flag is (has to be) set once a config has been read from EEPROM.
(One could choose to wait for a direct command to explicitly activate after reading - but why? And we would need a flag that ensures that a valid config has been read before activation - which is exactly the configActivated flag anyway).

elral:
Sorry for the time delay, but I am still busy in my daily work.

I am unhappy to have `resetDeviceConfig()` within the function `readStoredConfig()`. I think it should be part of `OnResetConfig()` as the reset parts should belong together and not one is part of a read/restore part.

Moving it back to `OnResetConfig()` I am wondering if `resetStoredConfig()` is still required, beside in my opinion resetting the config is not only setting the config lenght to 0. Clearing the device buffer belongs also to a resetted config.

GioCC:
Well, to my regret (I didn't think this could be such a time-consuming issue for everyone), I think I have to further comment.

Not having `resetDeviceConfig()` at the start of `readStoredConfig()` implies accepting that this latter is an inherently inconsistent (i.e. potentially dangerous) function: there is no scenario where it could be used without requiring `resetDeviceConfig()` to be first performed, and doing so inadvertently would cause a corrupt state. Therefore, sorry, I still can't see a valid reason why that statement should be incorporated there.

Resetting the EEPROM config (storage) and resetting the RAM config (working memory) are two different features. 
Currently, they are indeed used together - and this is not even true (loading the config at startup amounts to a `resetDeviceConfig()` alone, silently relying on the fact that at startup related variables are found nulled).
The reset parts "should" not belong together: they just happen to be naturally called together in the normal case.
But there might be different cases: I could think of custom firmwares where e.g. the memory config is initially read from EEPROM, then maybe altered according to some kind of inputs, and then cleared and re-read from the base config to be altered in a different way (config changes on-the-fly?).
I'm not saying that this could ever be a behaviour of the standard firmware (or any firmware for that matters), but IMO that's not a reason for the code not to be clean, correct and consistent.

Another (albeit not particularly nice) solution could be to group all these functions in a single one, which can't be used in a wrong way - probably the `OnResetConfig()` you think of. But this would unfortunately clash with the particular sequence of commands sent by the Connector for reconfiguring and reset (and with the existence itself of such commands). 

As for your last observation (having `OnResetConfig()` perform a complete cleanup and invalidate rather than just an EEPROM cleanup): this makes sense too, it's just a different interpretation of the specific command.
Wiping the stored config does not necessarily mean also wiping the working config before another one is available to replace it. But I can see the advantage that, by stopping the board, it makes clear that something is being changed (and in the meantime that the board will no longer operate if restarted), so probably your solution is preferable.

DocMoebiuz:
I really don't understand why we are putting so much effort into this change.

1. If i read the old code it is clear to me what it does.
2. When i read the new version I don't understand it right away.

Is there a existing or imminent bug that we are fixing? Are we really creating value with this change? Are we actually sure that the code is still exactly doing what it is supposed to?

I know @GioCC you are passionate about this, and I hate that I am not sold on it, but if we are not having any problems, why doing the change in the first place, especially if we feel that the readibility is not improved.

PR #198, opened 2022-10-02
(#197) Fix missing retrigger for Mux inputs at startup

### LINKED ISSUE #197: Retrigger doesn't work for multiplexers
https://github.dev/MobiFlight/MobiFlight-FirmwareSource/blob/5d1e90c821375ad134163542e1b5c957c75d1ba4/src/CommandMessenger.cpp#L104-L105

A call to OnTrigger() is missing from the above function

GioCC:
## Description of changes

Fixes #197

Included Mux input polling in the retrigger command

### DIFF
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -30,6 +30,9 @@
 #if MF_OUTPUT_SHIFTER_SUPPORT == 1
 #include "OutputShifter.h"
 #endif
+#if MF_DIGIN_MUX_SUPPORT == 1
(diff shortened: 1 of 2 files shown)


elral:
@GioCC @MobiFlight-Admin 
I am wondering if it would make sense to add here the initialisation of inputs according the actual status like described in issue #200 ? This would be comparable to PR #199 where the average buffer gets filled up on startup.

DocMoebiuz (approved):
Looking good to me. THANKS!

PR #199, opened 2022-10-02
(#898)(Connector) Fix missing retrigger for ANALOG inputs at startup

### LINKED ISSUE #202: Retrigger function does not initialize current value of analog potentiometers
Firmware copy of https://github.com/MobiFlight/MobiFlight-Connector/issues/898

**Describe the bug**
The retrigger function does not initialize the current value of analog potentiometers position.

**Steps To Reproduce**

Start Mobiflight with a potentiometer config assigned to a light dimmer in the sim and pot in mid position.
Press the retrigger button
Note lights in the sim did not sync to the pot position.
Expected behavior
Retrigger should synchronize the pot position in the sim.

**Actual behavior**
Potentiometer is ignored in the retrigger. Nothing happens.

**Details**

MobiFlight Version: 9.4.0.4
FlightSim: irrelevant
AddOn: irrelevant

**Additional context**
Workaround is to move the pot, which defeats the purpose of the retrigger function.

GioCC:
## Description of changes
Fixes #202 
Fixes #898 (listed under Connector issues)

Included Analog input polling in the retrigger command.
An initial scan has been added to setup() to fill the value buffer for analog inputs before it is used

### DIFF
--- a/src/MF_Analog/MFAnalog.cpp
+++ b/src/MF_Analog/MFAnalog.cpp
@@ -18,15 +18,25 @@ MFAnalog::MFAnalog(uint8_t pin, const char *name, uint8_t sensitivity)
     analogRead(_pin);            // turn on pullup resistors
 }
 
-void MFAnalog::update()
+void MFAnalog::readChannel(uint8_t compare)
 {
     int16_t newValue = ADC_Average_Total >> ADC_MAX_AVERAGE_LOG2;
-    if (abs(newValue - _lastValue) >= _sensitivity) {
+    if (!compare || abs(newValue - _lastValue) >= _sensitivity) {
         _lastValue = newValue;
         if (_handler != NULL) {
             (*_handler)(_lastValue, _pin, _name);
         }
     }
+}
+
+void MFAnalog::update()
+{
+    readChannel(true);
+}
+
+void MFAnalog::retrigger()
+{
+    readChannel(false);
 }
 
 void MFAnalog::readBuffer()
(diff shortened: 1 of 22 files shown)


elral:
I think initialisation of the average buffer should be part of the constructor `MFAnalog::MFAnalog` so that separate function is not required.
For my understanding all steps to get a class working should be part of the constructor if possible, if not it should be part of a `init()` function. As we have only the constructor and nothing else is required to get it working (e.g. SPI, serial, ...) I think it is the correct place to do it.

GioCC:
@elral you're absolutely right, thanks for noticing! I changed it the way you suggest,

elral:
@GioCC looks perfect now!

src/MF_Analog/MFAnalog.cpp
```diff
@@ -15,20 +15,32 @@ MFAnalog::MFAnalog(uint8_t pin, const char *name, uint8_t sensitivity)
...
     _lastValue   = 0;
     pinMode(_pin, INPUT_PULLUP); // set pin to input. Could use OUTPUT for analog, but shows the intention :-)
-    analogRead(_pin);            // turn on pullup resistors
+    for (uint8_t i = 0; i < ADC_MAX_AVERAGE; i++) {
+        readBuffer();
+    }
 }
 
-void MFAnalog::update()
+void MFAnalog::readChannel(uint8_t compare)
```
DocMoebiuz, line 23: the parameter name `compare` could be more explicitly describing what we mean so that it becomes easier to understand when reading:

`forceReadandAlwaysTrigger`
`alwaysTrigger`

And also 
```abs(newValue - _lastValue) >= _sensitivity```

would be nicer in a short function `valueHasChanged`

DocMoebiuz (changes requested):
please see my comment.

DocMoebiuz (approved):
Thanks @GioCC

PR #201, opened 2022-10-03
Improve getting correct device infos for connector from Mega

### LINKED ISSUE #200: Mega get's sometimes not properly identified
* eeprom config
* serial number
* devices

The same applies to:
* input shifter
* multiplexer
* analog in

elral:
## Description of changes
Buttons and input shifters are initialized with the actual status of inputs.
This avoids serial messages an startup due to changed status of inputs in case they are closed (ON).

Fixes #200

### DIFF
--- a/src/MF_Button/MFButton.cpp
+++ b/src/MF_Button/MFButton.cpp
@@ -12,8 +12,8 @@ MFButton::MFButton(uint8_t pin, const char *name)
 {
     _pin   = pin;
     _name  = name;
-    _state = 1;
-    pinMode(_pin, INPUT_PULLUP); // set pin to input
+    pinMode(_pin, INPUT_PULLUP);    // set pin to input
+    _state = digitalRead(_pin);     // initialize on actual status
 }
 
 void MFButton::update()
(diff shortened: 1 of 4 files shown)


DocMoebiuz (approved):
Thanks. That was fast.

GioCC:
Not really sure if duplicating the shift register poll in the constructor is nice; also it's not consistent with the approach used for analog inputs (use the existing function by adding an argument that says whether the change of value should trigger events).
I'd like to change that while adding the multiplexer part (that should complete this fix).

elral:
> Not really sure if duplicating the shift register poll in the constructor is nice; also it's not consistent with the approach used for analog inputs (use the existing function by adding an argument that says whether the change of value should trigger events). I'd like to change that while adding the multiplexer part (that should complete this fix).

Yeah, maybe you are right that doubling is not the best solution. I thought about this but came to the conclusion that it is less code which gets doubled compared to readability. And looking on all functions it is tripled (I didn't want to change the `update()` function).
I will have a closer look if/how it can be simplified/harmonized.

GioCC:
I have included those changes with the multiplexer mods that I will be posting shortly, you can have a look at them and see if they suit you.

Issue #159, opened 2022-02-24
Introduce an official setting for the VS Code C/C++ autoformatter 


GioCC:
In order to have a consistent code formatting, avoiding the recurring discrepancies in commits, we could conveniently use the automatic formatter of the standard "C/C++" plugin of VSCode, with a project-based setting. 

This could be accomplished by having a `settings.json` file saved in the `.vscode` directory of the project; attached is a sample with my current settings.

Reformatting can be made effortlessly by explicit user action, and also automatically upon file save.
I'm not really sure about this last option; it might confuse the user, but it would be a useful safeguard.

[settings.json.txt](https://github.com/MobiFlight/MobiFlight-FirmwareSource/files/8132191/settings.json.txt)

GioCC:
Link to [CLang formatting options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)

elral:
For me it is important that we all have the same settings, however they are. I am not very familiar with these settings, so I will use what every body else thinks is sensefull.

elral:
I tried this on my firmware for the motorized Trimwheel/Throttle and was wondering, that nothing has changed. Now I found with Autosave enabled you have still to actively save the file for reformatting.

I tested the settings and I am fine with them, would be good to have them in the near future for the next changes ;)

GioCC:
> /edit: It seems not to work with .h files, is this intended?
Hmmm, never noticed that... not 100% sure, but I'd swear it works for me... I'll check.
 
> /edit2: Oh, and these settings change the order of the `#include` files ...
Yes, that's something I'd like to rectify too. There's obviously a rule to it, but changing the order of includes might break compilation at times. Will look for that too..

elral:
Do we have the common understanding that the merged file with the autoformatter settings are agreed?
What about the .h files, seems still not to work for me after a short test.

DocMoebiuz:
Important to agree on this @GioCC and @elral.

GioCC:
@DocMoebiuz @elral @neilenns Guys, do you think we can close this issue considering the file already in the repo as accepted? Are there any other questions/issues?

elral:
For me it is fine, I checked also that the .h files are working for me.
However, this will us not protect against having changes on not changed code parts. If we add e.g. `#define EXAMPLE_TEST 200` to a `#define EXISTING 66`, then `#define EXISTING` gets filled up with spaces.
This will happen also on other parts.

[...]

PR #203, opened 2022-10-04
Use same retrigger logic for Multiplexer and InputShifter

### LINKED ISSUE #200: Mega get's sometimes not properly identified

GioCC:
Changes:

- [x] Added value initialization for digital input multiplexers
- [x] Introduce enum DO_TRIGGER and DONT_TRIGGER for both device types
- [x] Introduce poll method for both device types to reduce duplicate code
- [x] use same trigger method logic 

## Description of changes
relates to #201 
fixes #200

### DIFF
--- a/src/MF_InputShifter/MFInputShifter.cpp
+++ b/src/MF_InputShifter/MFInputShifter.cpp
@@ -1,4 +1,4 @@
-// 
+//
 // MFInputShifter.cpp
 //
 // (C) MobiFlight Project 2022
@@ -10,32 +10,27 @@ inputShifterEvent MFInputShifter::_inputHandler = NULL;
 
 MFInputShifter::MFInputShifter(const char *name)
 {
-  _initialized = false;
-  _name = name;
+    _initialized = false;
+    _name        = name;
 }
 
 // Registers a new input shifter and configures the clock, data and latch pins as well
 // as the number of modules to read from.
 void MFInputShifter::attach(uint8_t latchPin, uint8_t clockPin, uint8_t dataPin, uint8_t moduleCount, const char *name)
 {
-  _latchPin = latchPin;
-  _clockPin = clockPin;
-  _dataPin = dataPin;
-  _name = name;
-  _moduleCount = moduleCount;
+    _latchPin    = latchPin;
+    _clockPin    = clockPin;
+    _dataPin     = dataPin;
+    _name        = name;
+    _moduleCount = moduleCount;
 
-  pinMode(_latchPin, OUTPUT);
-  pinMode(_clockPin, OUTPUT);
-  pinMode(_dataPin, INPUT);
-  _initialized = true;
+    pinMode(_latchPin, OUTPUT);
+    pinMode(_clockPin, OUTPUT);
+    pinMode(_dataPin, INPUT);
+    _initialized = true;
 
-// And now initialize all buttons with the actual status
-  digitalWrite(_clockPin, HIGH); // Preset clock to retrieve first bit
-  digitalWrite(_latchPin, HIGH); // Disable input latching and enable shifting
-  for (int module = 0; module < _moduleCount; module++) {
-    _lastState[module] = shiftIn(_dataPin, _clockPin, MSBFIRST);
-  }
-  digitalWrite(_latchPin, LOW); // disable shifting and enable input latching
+    // And now initialize all buttons with the actual status
+    poll(DONT_TRIGGER);
 }
 
 // Reads the values from the attached modules, compares them to the previously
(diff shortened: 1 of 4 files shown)


src/MF_InputShifter/MFInputShifter.cpp
```diff
@@ -10,146 +10,136 @@ inputShifterEvent MFInputShifter::_inputHandler = NULL;
...
 }
 
 void MFInputShifter::detach()
 {
-  if (!_initialized)
-    return;
-  _initialized = false;
+    if (!_initialized)
+        return;
+    _initialized = false;
```
elral, line 142: Isn't just `_initialized=false;` enough?

src/MF_InputShifter/MFInputShifter.cpp
```diff
@@ -10,146 +10,136 @@ inputShifterEvent MFInputShifter::_inputHandler = NULL;
...
 {
     if (state == LOW && _inputHandler != NULL) {
-    (*_inputHandler)(inputShifterOnPress, pin, _name);
+        (*_inputHandler)(inputShifterOnPress, pin, _name);
     } else if (_inputHandler != NULL) {
-    (*_inputHandler)(inputShifterOnRelease, pin, _name);
-  }
+        (*_inputHandler)(inputShifterOnRelease, pin, _name);
+    }
 }
```
elral, line 131: Would it make sense to use the same code as for the multiplexer?
`if (!_inputHandler) return;
(*_inputHandler)((state ? inputShifterOnRelease : inputShifterOnPress), pin, _name);`

elral:
Additionally we have still the topic with auto formatting the code.
I still don't know if something has to be add to my personal platformIO or if a general setting has to be released.

src/MF_InputShifter/MFInputShifter.cpp
GioCC: Yes it would. Didn't go over this section of code.

GioCC:
> 



> Additionally we have still the topic with auto formatting the code. I still don't know if something has to be add to my personal platformIO or if a general setting has to be released.

I believe I'm using the last version that was basically agreed upon; I don't recall if it was ever been made official by incorporating it in the project (but I guess we both would have found it in the git checkout if this was the case).

DocMoebiuz:
> > 
> 
> > Additionally we have still the topic with auto formatting the code. I still don't know if something has to be add to my personal platformIO or if a general setting has to be released.
> 
> I believe I'm using the last version that was basically agreed upon; I don't recall if it was ever been made official by incorporating it in the project (but I guess we both would have found it in the git checkout if this was the case).

you have to check indentation settings, the code used to be 2 spaces per indent, your settings seems to be 4.

GioCC:
> > > 
> > 
> > 
> > > Additionally we have still the topic with auto formatting the code. I still don't know if something has to be add to my personal platformIO or if a general setting has to be released.
> > 
> > 
> > I believe I'm using the last version that was basically agreed upon; I don't recall if it was ever been made official by incorporating it in the project (but I guess we both would have found it in the git checkout if this was the case).
> 
> you have to check indentation settings, the code used to be 2 spaces per indent, your settings seems to be 4. Maybe I am wrong, I looked at the code... majority is now 4 spaces.
> 
> I looked back in history, it used to be 2 spaces only. So yes, we have to have a setting that we are all using. Which file is it? Maybe it is not committed to the repository?
> 
> "editor.formatOnSave": true - this is also the reason that we see the format changing.

https://github.com/MobiFlight/MobiFlight-FirmwareSource/blob/a57611caa55660e2459c5825715226c82dab1598/.vscode/settings.json#L2

elral:
> It's there, so I assume it was agreed upon at some point. Anyway it's what affects my saves.

Ah, yes, I now remember. It was agreed on this, but I also remember that this does not "work" for autosave which I have enabled (and forgot completly). I have to actively save the file. I will do my best to do it before commiting new files...

elral:
I just doubled checked the formatting, and indeed the file will not be formatted if it is not actively saved (CTRL-S or from menu).
As an example see the attached screen shot. Only with pressing CTRL-S the file is formatted what I havn't done before because I have Autosave enabled. Sorry for this.
But I am still wondering why in this file the TabWidth is still 2 and not 4 according the settings.

image

GioCC:
> I just doubled checked the formatting, and indeed the file will not be formatted if it is not actively saved (CTRL-S or from menu). As an example see the attached screen shot. Only with pressing CTRL-S the file is formatted what I havn't done before because I have Autosave enabled. Sorry for this. 

Thinking about it, it totally makes sense, because otherwise autosave would possibly reformat your code "under your feet" while you are working... (I personally refrain from using autosave, so I never noticed)

> But I am still wondering why in this file the TabWidth is still 2 and not 4 according the settings.

This file had probably not been modified, thus saved, in a long time.

I think it would make sense to make a specific commit only with reformatting applied to all files, that would avoid recurring problems.
 
> /edit: sometimes it is good to check all open issues -> #159 ;) We should continue the discussion of formatting in this issue...

Yes. Often small (meaning, quick to solve) issues linger around for long.
Though the presence of the config file seems to suggest that the issue was decided at some point in time?

neilenns:
This is the solution to making format happen on auto save: https://marketplace.visualstudio.com/items?itemName=pucelle.run-on-save

Vs code doesn't support it natively despite many requests.


neilenns:
Here's a sample settings.json snippet for that extension:

```
"runOnSave.commands": [
    {
        "match": ".ps1",
        "command": "editor.action.formatDocument",
        "runIn": "vscode"
    }
]
```

neilenns:
Or try this: https://marketplace.visualstudio.com/items?itemName=BdSoftware.format-on-auto-save

DocMoebiuz:
> Or try this: https://marketplace.visualstudio.com/items?itemName=BdSoftware.format-on-auto-save

for me i have that setting in the cpp extension which is auto format on save but not auto format on auto save.

elral:
@neilenns thanks for the informations! I will give them a try.

elral:
> (Though the presence of the config file seems to suggest that the matter was decided at some point in time?)

At least it was merged.

DocMoebiuz:
@Gioc

> Fix: Added initial value setting also for digital input multiplexers Mod: Refactored duplicate code in input shifter handling
> 
> ## Description of changes
> Completed fixes for #200 as follow-up to PR #201

@GioCC please add some details about this change, so that it is easier for me to understand the motiviation, the improvement and the scope. What is the follow up on PR #201 exactly?

DocMoebiuz (changes requested):
Hi @GioCC thanks for helping with this. For an easier review, please consider the following points:

* Provide a clear and simple PR title that summarizes what the PR does, e.g. in this case
  * **Use same retrigger logic for Multiplexer and InputShifter**

* Provide a short outline of the significant changes, e.g. as a checkbox list
  - [ ] Introduce enum DO_TRIGGER and DONT_TRIGGER for both device types
  - [ ] Introduce poll method for both device types to reduce duplicate code
  - [ ] use same trigger method logic 

That would help me a lot to faster grasp the intention of the PR and reviewing the code.

DocMoebiuz (approved):
I made the changes for you in the comment.
Thanks @GioCC

PR #208, opened 2022-10-09
Increase number of multiplexer devices (Fixes #194)

### LINKED ISSUE #194: Increase the supported Mux for Mega to 3 devices
At the moment we can only add up to 2 16-bit Mux modules.
Tests have shown, that you can have 3 connected in a stable way.

So we can up increase the limit to 3 on the mega. Uno and Pro Micro can stay the same.

GioCC:
Fixes #194
Increases number of available multiplexer devices from 2-1-2 (Mega/ProMicro/Uno) to 4-3-3.


_Boards/Atmel/Board_ProMicro/MFBoards.h
```diff
@@ -17,14 +17,15 @@
 #define MF_STEPPER_SUPPORT   1
 #endif
 #ifndef MF_SERVO_SUPPORT
-#define MF_SERVO_SUPPORT 1
+#define MF_SERVO_SUPPORT     0
```
neilenns, line 20: Was it intentional to remove servo support from the pro micro?

src/Config.cpp
```diff
@@ -12,25 +12,31 @@
 #include "Output.h"
```
neilenns, line 12: This file is all whitespace changes and is unrelated to changing the number of multiplexers. Probably shouldn't be included in this PR.

neilenns: Oh wait, I see it changes to include mux header files. Was that intentional?

_Boards/Atmel/Board_ProMicro/MFBoards.h
GioCC: No, this change (and other ones below) have been introduced by inadvertently basing on a local main branch that had strayed from the  official main. Sorry for the mishap, fixed now.

src/Config.cpp
GioCC: See other comment above.

neilenns:
@GioCC The rogue changes are still there. Since these changes are so small it might make sense to start from a fresh branch that is based on the current main. This PR has a lot of commits that are gonna get squashed...

GioCC:
> @GioCC The rogue changes are still there. Since these changes are so small it might make sense to start from a fresh branch that is based on the current main. This PR has a lot of commits that are gonna get squashed...

Right, I was just noticing that. I'm doing exactly as you suggest, but also deleting that pesky "main" branch later...

PR #209, opened 2022-10-09
Increase number of multiplexer devices (CORRECTED PR)

### LINKED ISSUE #194: Increase the supported Mux for Mega to 3 devices

GioCC:
Fixes #194

Number of multiplexer devices increased from 2-1-2 (Mega/ProMicro/Uno)
to 4-3-3.

### DIFF
--- a/_Boards/Atmel/Board_Mega/MFBoards.h
+++ b/_Boards/Atmel/Board_Mega/MFBoards.h
@@ -8,19 +8,19 @@
 #define MFBoardMega_h
 
 #ifndef MF_SEGMENT_SUPPORT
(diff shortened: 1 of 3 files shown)


neilenns:
Nice!

DocMoebiuz (approved):
Thanks.

@GioCC  Please check your auto format on save settings because there are many lines changed that actually are no change, only indentation.

PR #212, opened 2022-10-12
Fix faulty initialization of analog value at startup

### LINKED ISSUE #200: Mega get's sometimes not properly identified

GioCC:
## Description of changes

Fixes #200 (reopened) but belongs to the scope of the changes made in PR #199.

- [X] Added initialization of actual value returned, as opposed as buffer values only, in the constructor.


elral:
That's what I also prepared, thanks again for double working.

DocMoebiuz (approved):
Alright! Cool. Thanks!

PR #215, opened 2022-10-21
Custom device additions

elral:
This is the second part of the pull request for the custom devices.
It is also foreseen to have a basis for further discussions.
It deals with additional functions for some existing devices which coudl be sensefull to use within a custom device.
Additionally each device reports back its registered number to have directly access to this device from the custom device without the need that it gets registered from the connector. The required pins (if there are some) must be in this case defined within the connector dialog for the custom device.

As an example that these functions could be sensefull have a look on my example for a custom device which sets up a motorized axis. It is under [MotAxis](https://github.com/elral/MobiFlight-FirmwareSource/tree/Custom_Device_MotAxis).


DocMoebiuz:
deprecated

PR #217, opened 2022-10-24
MF type to flash, simplier check for serial avail.

### LINKED ISSUE #216: Small RAM saving
For now `MOBIFLIGHT_TYPE`  is stored in the RAM via `const char type[sizeof(MOBIFLIGHT_TYPE)] = MOBIFLIGHT_TYPE;`.
The type is only send to the connector and not used anymore. So it could be stored in the flash.

In `void restoreName()` a char array of 1 is defined which is not needed, MFeeprom.read_char() could be used.

elral:
Store `MOBIFLIGHT_TYPE` in flash and not in RAM.

Easier function to check if a name is available.

Fixes #216

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -461,8 +461,7 @@ void OnGenNewSerial()
 // ************************************************************
 void storeName()
 {
-    char prefix[] = "#";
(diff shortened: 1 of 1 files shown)


DocMoebiuz (approved):
YEAH!

PR #219, opened 2022-11-09
send deviceID instead of name from input device

### LINKED ISSUE #207: Eliminate the nameBuffer for freeing up additional RAM
As discussed on Discord already the nameBuffer could be eliminated and instead of the name from an input device an "identifier" would be send. The connector have to know about the identifier and the relation to the input device name, therefore the user will not seen any differences. The benefit is that the nameBuffer stored in the RAM is not required anymore and additional RAM is available. Furthermore the serial traffic is reduced as the names could be "long" and instead just an identifier is transferred.

The identifier could be the x.th device from the same type which gets initialised. This information is already available from number of intialised decices, e.g. for the buttons `buttonsRegistered` in `buttons[buttonsRegistered] = new (allocateMemory(sizeof(MFButton))) MFButton(pin, name);`. Instead of saving the name `buttonsRegistered` would be saved and on an event transferred.
During processing the config there is always the same order of initialising input devices so there is also always the same `indentifier`.
Reading the config from the connector happens also in the same order, therefore the connector has a fix relation between the `identifier` (x.te device) and the name which is transferred during reading the config.
The same applies to all other input devices.

For backwards capability the connector has to check the firmware version and depending on this the connector has to expect names or identifiers.

elral:
## Description of changes

All input devices will now send there deviceID (means the xth device of same type registered) to the connector instead of the name if the status has changed.
This will reduce the serial traffic and reduce the RAM consumption as the NameBuffer is not required anymore.

@GioCC will issue an additional pull request with the required changes for the input multiplexer. These changes are not part of this pull request. GioCC will also add the comment that issue 207 will be fixed, so it's not required here.

Changes in the connector are also required as the names are not send anymore.

Fixes #207 

/edit: added reference to issue


GioCC:
@elral Sorry, I forgot that file.  
The offending `MuxDigInEvent` typedef declaration __has to be completely removed from `MFMuxDriver.h`__, it had no business being there in the first place (as it's already in `MFDigInMux.h` where it belongs).

elral:
There is no real demand for now to implement this feature. Also with all the memory optimazations from the past RAM is no real bottleneck anymore. We are short for some boards, but it's working. Additional features are also limited by available flash for some boards.
For future features it might also be sensefull to send the names instead of the deviceID.

PR #222, opened 2022-11-23
add write block w/o length, renaming read_char

### LINKED ISSUE #221: EEPROM  optimisation
There could be some smaller optimisations in the EEPROM class.
Mostly for reading from EEPROM a byte is required, but we are reading a char.
Reading/saving a complete array is not foreseen, but might be required for future use (e.g. reading/saving a struct).

elral:
`read_char()` is renamed to `read_byte()` as mainly a byte is required.
Within `OnGetInfo()` a char is required to send the config via the command messenger, at this point `read_byte()` is casted.
Additional functions to save/read a complete array or struct. This could be helpfull for future use (e.g. parameters for an analog calibration should be in a struct and this struct should be saved to the EEPROM). These additional functions and the existing functions for reading/saving arrays with fixed lenght are moved to `EEPROM.h` in a template.

`storeName()` and `restoreName()` are cleaned up slightly

Fixes #221

### DIFF
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -78,7 +78,7 @@ void attachCommandCallbacks()
     cmdMessenger.attach(kSetLcdDisplayI2C, LCDDisplay::OnSet);
 #endif
 
-#if MF_OUTPUT_SHIFTER_SUPPORT
+#if MF_OUTPUT_SHIFTER_SUPPORT == 1
     cmdMessenger.attach(kSetShiftRegisterPins, OutputShifter::OnSet);
 #endif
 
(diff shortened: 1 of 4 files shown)


DocMoebiuz (approved):
TY!

GioCC:
Not sure about the benefit of templating these functions.
I'm afraid that the compiler might generate redundant instances of the same function (unless we trust it to be smart enough - which I admit it might well be) if the template is instantiated for different types.
I see that even the basic EEPROM.get() uses templating - but it's still a framework library, so redefining it would get in the way of porting.
Passing <t> by reference has no particular purpose other than mask the fact that we are getting its raw pointer anyway. What's bad in plainly invoking the basic functions with & and sizeof()?
Besides, defining the templates in the .h file rather than in the .cpp should inline them - is that desired?

(BTW, casting to uint8_t* would not work for non-POD objects - which we are not currently using and won't probably be using anytime soon, but still something to be aware of.)

PR #227, opened 2022-12-19
faulty init on startup, check for available memory

### LINKED ISSUE #226: Input Multiplexer get not initialized to current status on start up
We had  some pull request to solve issue #200 but the input multiplexer gets still not initialized to the start up status in the end.

Furthermore it is not checked if adding a DigInMux fits into the device buffer like the other devices.

elral:
On startup the constructor for the digital input multiplexer is called and the status of each input gets initialized to the current status. With one of the next steps this status get be cleared again, so on startup serial messages could be generated due to status changes of the inputs.
Furthermore it is not checked like for the other devices if the digital input multiplexer fits into the device buffer.

Calling the clear() function is deleted.
Check if device fits into the device buffer is added.

Fixes #226

### DIFF
--- a/src/MF_DigInMux/DigInMux.cpp
+++ b/src/MF_DigInMux/DigInMux.cpp
@@ -29,11 +29,14 @@ namespace DigInMux
         if (digInMuxRegistered == MAX_DIGIN_MUX)
             return;
         MFDigInMux *dip;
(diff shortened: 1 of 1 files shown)


DocMoebiuz (approved):
Thanks! Good catch.

PR #228, opened 2022-12-20
Pico

### LINKED ISSUE #225: Support the Raspberry Pi Pico

elral:
This pull request adds the support for the Raspberry Pi Pico.
The supported board is the original board from the Raspberry Pi Foundation.

* A environment is added to the platformio.ini file
* A MFBoards.h file is added for the board specific definitions
* After writing to the eeprom these changes have to be commited, accordingly MFEEPROM.cpp/h is modified
* Pico requires it's own servo lib. Therefore the servo lib for Atmels is moved from the standard environment to the boards environment from the Atmel boards
* The Pico servo lib gets initialized with the same range as for the Atmels
* The unique ID from the NOR flash gets used instead of a random serial number
* The device buffer gets initialized with the correct size for this (and other) architectures.

Fixes #225


neilenns:
YES!!!!!!!!!!!!!!!!!!!!!!!!!!!!

_Boards/RaspberryPi/Pico/raspberry_pico.board.json
```diff
@@ -0,0 +1,203 @@
+{
```
neilenns, line 1: We haven't checked the .board.json file into this github repo for any of the amtel platforms so this should probably be excluded from this PR.

platformio.ini
```diff
@@ -61,7 +60,8 @@ build_src_filter =
 	${env.build_src_filter}
 	+<../_Boards/Atmel>
 lib_deps = 
-	${env.lib_deps}	
+	${env.lib_deps}
+	arduino-libraries/Servo @ 1.1.8
```
neilenns, line 64: Instead of putting the arduino library in each arduino env, how about creating a new `lib_amtel_deps` section then including that as `${env.lib_amtel_deps}`?

neilenns (changes requested):
So exciting! Two minor suggestions. Looks great!

_Boards/RaspberryPi/Pico/raspberry_pico.board.json
elral: Ups, you are rigth. Was the best location for me during delelpment.
I will delete it.

platformio.ini
elral: Ah, very good idea. I will change this.

neilenns: Maybe pair it with a `lib_rp2040_deps` too, even though there's just one rp2040 platform?

elral: Hmhm, there are no additional libs for the Pico.

neilenns (approved):
It's happening!!! YES!!!

PR #230, opened 2022-12-28
Unique

### LINKED ISSUE #229: Sometimes the same serial number gets generated
For a "real" random number a random starting point is required. It is coming practise and also used in the firmware to use an floating analog input which gives some kind of random number.
The problem is, once an input device is connected to the analog input this starting point will always be the same and therefore always the same random number will be generated.
If this input device is e.g. button, the analog value will be 0 or 1023 also on different Arduinos. In this case the same serial number on different Arduinos will be generated.

elral:
@DocMoebiuz and @neilenns found a library with an undocumented feature to read an UniqueID on Atmega2560 and ATmega328. For new boards on first start up this UniqueID is used. In the library is also stated, that it might be possible that
two or more could have the same UniqueID (https://github.com/ricaun/ArduinoUniqueID/blob/master/MCU.md#Disclaimer).
My tests have shown that my megas have different UniqueID's.
But as this can not be excluded, generating a serial number as command from the connector is still available. In this case the random generator is used as before but the starting point are the milliseconds from starting up the board. It is very! unlikely to generate the serial number on exactly the same milliseconds between different boards, so a double serial number should not be generates anymore.

On startup of the board it is checked if a serial number was always generated. This is marked in the EEPROM with "SN" on the first two memory locations. In this case this serial number will still be used. Nothing changes for the user.

If a serial number is not available, it is checked if "ID" was written to the EEPROM. This means it is NOT the first start up and the UniqueID will be used as serial number.

If "SN" and "ID" is not read from the EEPROM, it is a first start up of the board an the UniqueID is used as serial number. This is also marked with writing "ID" to the first two EEPROM locations from the serial number locations.

If the user actively request a new serial number, a serial number according the existing style is generated and marked in the EEPROM with "SN". The starting point for the random generator is NOT anymore the analog input A0, it is the time difference from starting up the board until the command to generate a new serial (simply millis()).

In case a UniqueID is used, the serial number is longer than before. The array size for storing the serial number is calculated accordingly for the max. numbers of characters. The original definition `MEM_LEN_SERIAL` is kept, otherwise the device configuration would start at another position in the EEPROM.

Fixes #229

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -6,10 +6,10 @@
 
 #include "mobiflight.h"
 #include "MFEEPROM.h"
-
 #include "Button.h"
 #include "Encoder.h"
 #include "Output.h"
+#include "ArduinoUniqueID.h"
 
 #if MF_ANALOG_SUPPORT == 1
 #include "Analog.h"
@@ -55,12 +55,12 @@ const uint8_t MEM_OFFSET_SERIAL = MEM_OFFSET_NAME + MEM_LEN_NAME;
 const uint8_t MEM_LEN_SERIAL    = 11;
 const uint8_t MEM_OFFSET_CONFIG = MEM_OFFSET_NAME + MEM_LEN_NAME + MEM_LEN_SERIAL;
 
-char      serial[MEM_LEN_SERIAL]     = MOBIFLIGHT_SERIAL;
-char      name[MEM_LEN_NAME]         = MOBIFLIGHT_NAME;
-const int MEM_LEN_CONFIG             = MEMLEN_CONFIG;
-char      nameBuffer[MEM_LEN_CONFIG] = "";
-uint16_t  configLength               = 0;
-boolean   configActivated            = false;
+char      serial[UniqueIDsize * 2 + 1] = MOBIFLIGHT_SERIAL;
+char      name[MEM_LEN_NAME]           = MOBIFLIGHT_NAME;
+const int MEM_LEN_CONFIG               = MEMLEN_CONFIG;
+char      nameBuffer[MEM_LEN_CONFIG]   = "";
+uint16_t  configLength                 = 0;
+boolean   configActivated              = false;
 
 void resetConfig();
 void readConfig();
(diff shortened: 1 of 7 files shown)


src/Config.cpp
```diff
@@ -437,16 +437,52 @@ bool getStatusConfig()
 // ************************************************************
 void generateSerial(bool force)
 {
-    MFeeprom.read_block(MEM_OFFSET_SERIAL, serial, MEM_LEN_SERIAL);
-    if (!force && serial[0] == 'S' && serial[1] == 'N')
+    // A serial number according old style is already generated and saved to the eeprom
+    // So keep it to avoid a connector message with orphaned board
+    if (!force && MFeeprom.read_byte(MEM_OFFSET_SERIAL) == 'S' && MFeeprom.read_byte(MEM_OFFSET_SERIAL + 1) == 'N') {
```
neilenns, line 442: All of the if statements in this method start with `!force`, which should get refactored out into a standalone `if (force)` test at the start of the function to do whatever force does (which as far as I can tell is nothing?). This will make for easier to follow function logic and simpler conditions on the rest of the if statements.

src/Config.cpp
```diff
@@ -437,16 +437,52 @@ bool getStatusConfig()
...
+            if (UniqueID[i] < 0x10) {
+                sprintf(&serial[3 + i * 2], "0%X", UniqueID[i]);
+            } else {
+                sprintf(&serial[3 + i * 2], "%X", UniqueID[i]);
+            }
+        }
+        // and mark this in the eeprom
+        MFeeprom.write_block(MEM_OFFSET_SERIAL, "ID", 2);
+        // Set first byte of config to 0x00 to ensure with empty config on 1st start up
+        MFeeprom.write_byte(MEM_OFFSET_CONFIG, 0x00);
```
neilenns, line 474: The only difference between these two blocs is whether the word "ID" gets written to eeprom. There's no real downside to writing "ID" every time this is called.

Combined with my prior comment about `force` being tested first I believe that would simplify this function down to three cases:

1. force is true
2. SN is in eeprom so there's an old-style ID on the board
3. all other cases

src/mobiflight.cpp
```diff
@@ -9,6 +9,7 @@
 #include "Button.h"
 #include "Encoder.h"
 #include "MFEEPROM.h"
+#include "ArduinoUniqueID.h"
```
neilenns, line 12: I don't think this needs to be included in mobiflight.cpp since `UniqueId` isn't used anywhere in the file.

src/Config.cpp
neilenns, line 474: The only difference between these two blocks is whether the word "ID" gets written to eeprom. There's no real downside to writing "ID" every time this is called so this could be simplified significantly and an entire if statement could be removed.

Combined with my prior comment about force being tested first I believe that would simplify this function down to three cases:

1. force is true
2. SN is in eeprom so there's an old-style ID on the board
3. all other cases

src/Config.cpp
neilenns, line 442: All of the if statements in this method start with !force, which should get refactored out into a standalone if (force) test at the start of the function to do whatever force does (which as far as I can tell is nothing?). This will make for easier to follow function logic and simpler conditions on the rest of the if statements.

neilenns (changes requested):

src/mobiflight.cpp
elral: Oh yes, that's an fragment from my testing.

src/Config.cpp
elral: Let me think about this. I will avoid writing to the EEPROM every time the board is started up (so just for reading the EEPROM).

src/Config.cpp
elral: Yeah, good hint!

src/Config.cpp
neilenns: I think the function can simplify down to something like this:

```C++
if (!force && legacy value stored)
{
    read legacy value;
    return;
}

read unique ID;

if (unique ID marker isn't stored)
{
   store ID marker in eeprom;
}
```

I'm pretty sure this covers all the cases. It will only write ID once, and will allow someone to "generate serial" from the desktop to get a new-style ID.

elral:
I still have to test the ProMicro and how to handle the UniqueID if it works (not listed as processor w/o UniqueID).

neilenns (approved):
Love the updated version!

DocMoebiuz (approved):
Nice... this will hopefully help with this weird problem.

PR #231, opened 2022-12-29
Make the 7segs more reliable by resetting before a new display

peter-masters:
## Description of changes

The 7 segment displays do not always show updates properly.  After time, they become less reliable because they are never reset (shutdown).  I found that this is easily resolved by including a shutdown, reset the scan limit (to 7, the max to ensure uniform brightness) and re-activation on each display call.

According to the datasheet, the MAX7219 chip initialises the scan limit to 1 initially, so this also ensures it is correctly set.
https://www.sparkfun.com/datasheets/Components/General/COM-09622-MAX7219-MAX7221.pdf 

Pull request notes:
Sorry, this is my first pull request to you guys, so I hope this is ok, and happy to discuss.  My email is peter.masters at gmail dot com.  I tested this by running the mobiflight "test" cycle against 4 displays, which corrupt after some minutes (which I could test repeatedly).  When I ran a separate arduino program to test the digits, this never caused the issue.  By working through the logic, I found the difference compared with the LedControl logic you've used, and tried out the difference.  With this fix in place, the 7 segment displays remain reliable indefinitely.
_Describe changes here_

### DIFF
--- a/src/MF_Segment/MFSegments.cpp
+++ b/src/MF_Segment/MFSegments.cpp
@@ -17,6 +17,9 @@ void MFSegments::display(byte module, char *string, byte points, byte mask, bool
         return;
     byte digit = 8;
     byte pos   = 0;
+    _ledControl.shutdown(module, true);
+    _ledControl.setScanLimit(module, 7);
+
     for (uint8_t i = 0; i < 8; i++) {
         digit--;
         if (((1 << digit) & mask) == 0)
@@ -24,6 +27,8 @@ void MFSegments::display(byte module, char *string, byte points, byte mask, bool
         _ledControl.setChar(module, digit, string[pos], ((1 << digit) & points));
         pos++;
     }
+
+    _ledControl.shutdown(module, false);
 }
 
 void MFSegments::setBrightness(byte module, byte value)


elral:
> Sorry, this is my first pull request to you guys, so I hope this is ok, and happy to discuss. My email is peter.masters at gmail dot com. I tested this by running the mobiflight "test" cycle against 4 displays, which corrupt after some minutes (which I could test repeatedly). When I ran a separate arduino program to test the digits, this never caused the issue. By working through the logic, I found the difference compared with the LedControl logic you've used, and tried out the difference. With this fix in place, the 7 segment displays remain reliable indefinitely.

That's fine, but normally we start with an issue where we describe the problem/improvement and within the issue we describe the solution.

Regarding the prposed changes it migth be good if also @GioCC can have a look on this as he is quite familiar with the MAX7219 chips.

GioCC:
@peter-masters @elral If I should place my bet, I would guess that these issues arise from power supply problems (filtering or inadequate supply), which cause IC reset or lock-out.
On the other hand, MAXes notoriously have little tolerance for imperfect supplies or driving signals, and since less than ideal conditions are unfortunately to be expected in sim-cockpit installations (especially from inexperienced users), this modification makes sense to me.
BTW, at times I, too, added this kind of periodic reset sequences in the past, when there were no strict time constraints.

I wouldn't normally be very happy with adding redundant commands which - particularly on long chains - will add a non negligible time to the polling loop; however, given the necessity, I don't see any clearly better solution - I subscribe to the comment @elral made in its /edit.

Regarding the choice of sending data in shutdown mode: I fail to see a real necessity for it - however, the shutdown could at least be beneficial for the reset part, therefore there should be at least no particular harm in extending it to the digit setting part (unless the longer blanking time becomes more noticeable and annoying).

peter-masters:
Thanks both.  Sorry for not raising an issue, I should have thought to do that honestly!

I suspect @GioCC is correct in that at some level, there is still power issues in my setup.  I've spent Christmas trying to focus on it.  In my case, I have 4 x 7-seg displays running off a mega2560.  Each have power run in parallel to each board (as everyone suggests) and whilst originally I had a 6v power source (I was hoping this compensated for some of the under-voltage issues people seemed to have) I ultimately used a 1.5a 5v regulator and it's pretty solid).  One thing I did though as part of this was to make my own arduino code to drive those displays, and I noticed that I could run my code for hours (it just cycles through 11111111,22222222 etc at a choice of speeds) whereas the mobiflight code would corrupt the displays in seconds or minutes.  This then caused me to do a deep dive on the mobiflight firmware and proving out what I did.  With this fix, the displays in mobiflight are rock solid.  

I suspect on a good day, it might be fine, but as you say, I think this just provides that extra level of stability.  I don't think my setup is particularly weird, and I'm using short cables and not even daisy chaining the displays at this point, so in that sense, it'd only get worse if I did.  This fix is really about proposing a periodic stability fix which just forces the displays to maintain the right state.  I ran the code at a reasonably high refresh rate (went down to 80ms), and I don't see any downsides even for a heavy update speed, so I think this should hopefully help some people.  I couldn't see an obvious other point to call this (though I guess you could choose to do it every so many refreshes) but as I said, I don't think it causes issues with blanking that I can detect.

peter-masters:
Here's the test code I wrote that uses the LedControl library as per mobiflight with 4 displays.  Just tested with a 10ms 'speed' in the loop and it seems fine:

#include <LedControl.h>

//                  DIN CLK  CS
LedControl altitude( 41, 37, 39);
LedControl heading(  29, 25, 27);
LedControl crs_mach( 31, 35, 33);
LedControl vertspeed(47, 45, 43);

void setup() {
  altitude.setIntensity(0,8);
  heading.setIntensity(0,8);
  crs_mach.setIntensity(0,8);
  vertspeed.setIntensity(0,8);

  altitude.shutdown(0,true);
  heading.shutdown(0,true);
  crs_mach.shutdown(0,true);
  vertspeed.shutdown(0,true);
}

void loop() {
    int speed = 50;
    displayString(altitude,  "11111111");
    displayString(heading,   "22222222");
    displayString(crs_mach,  "33333333");
    delay(speed);
    displayString(altitude,  "22222222");
    displayString(heading,   "33333333");
    displayString(vertspeed, "11111111");
    delay(speed);
    displayString(altitude,  "33333333");
    displayString(crs_mach,  "11111111");
    displayString(vertspeed, "22222222");
    delay(speed);
    displayString(heading,   "11111111");
    displayString(crs_mach,  "22222222");
    displayString(vertspeed, "33333333");
    delay(speed);
}

void displayString(const LedControl& display, char* value) {
  display.shutdown(0,true);   // My fix
  display.setScanLimit(0,7);  // My fix
  
  for (int i=0; i < 8; i++) {
    display.setChar(0,i,value[i],false);
  } 
  display.shutdown(0,false);  // My fix
}

elral:
@peter-masters I wanted to give a try to your test code, but couldn't get it to run. I am wondering about the initialization of the 7Segment, as I can see the library has no constructor with parameters but instead is using a `begin()` function.

So first I made a modification to the original code and added a function which is writing continously in a loop to two daisy chained displays. This function is called after adding these both displays (called from config.cpp). This means that the everything else is blocked as I am in the continously loop. In this configuration my display worked over hours.

Next I changed this that the test display function (sending 0 to 9 for every digit) is called from the main loop every 100ms so that all other functions still work. In this case the display stopped working (just blank) within 1 hour (could be earlier as I leaved it running and checked after one hour). A second attempt shows that the display was working ~20min. before it stopped working (again just a blank display).

I will do another test according the first modification where it is just writing to the display in the continously loop w/o the other functions from mobiflight.

@GioCC Regarding a reset on a regularly basis I am thinking about an additional function which is called from the main loop every xyz minutes (tbd) like the input devices.

elral:
Another idea would be to check inside the 7Segment class for xyz minutes if a reset is required. If so than reset the display and afterwords send the characters to the display.
If it is done from the main loop I am afraid that the reset function is called, the display gets cleared and it takes some time to receive new data for the display. In this timeframe the display would be blank...

neilenns:
> /edit: I forgot to mention that my cabling is also not the best in my test configuration

Sounds like it's a valid real-world configuration then, since the cabling is rarely the best for users either 😂

elral:
The blank display came from the power saving, so my fault (thanks to @MobiFlight-Admin for the hint) . After deactivation of the power saving the segments were also running without problems for hours.
According @MobiFlight-Admin less/no issues are reported in the past with correct cabling.

peter-masters:
Thanks @elral ,
Just a note on the constructor difference.  This seems to be a diff in the mobiflight version.  I did notice this - it looks like the mobiflight version is modified from the original in that the constructor has been changed to a no-op constructor with no parms, and the functionality moved into a begin() method.  In my test, I was using the vanilla arduino IDE so I pulled down the original LedControl library via that IDE from the original source (same version of library) which has a parameterised constructor (https://github.com/wayoda/LedControl/blob/master/src/LedControl.h).  I assumed this was a deliberate need/modification in mobiflight to defer initialising the pins for the modules, but I guess comes with the concern that you're deviating from the main code branch of the dependency.  I guess its up to you guys as long as its a conscious decision :).

In terms of power issues - I certainly did have power issues in my setup (see this for details: https://www.mobiflight.com/forum/topic/7430.html) at that was me trying various things which ended up with a 6v supply instead of 5v (partly as I'd read about issues related to voltage drop but anyway, resolved by using a 5v regulator).  Anyway, it shows me setup and it really is just 4 displays running on 3 dedicated pins each of the arduino with separately supplied 5v power and a common ground.  I guess RF interference is always a possibility, but for me the fix I did was night & day compared with without.  I'm sure it can be improved on, though.  In theory the power saving shouldn't kick in with normal operation as you get regular pings that reset the timeout, but probably not if you blocked that out with hard-looping in the code for your test as you described.

elral:
@peter-masters 
> Just a note on the constructor difference. This seems to be a diff in the mobiflight version.

Oh yeah, you are completly right, and even it was me who did this modification. Just forgot this. The reason is that a separate `begin()` function needs less RAM than to have it in the constructor. And we had to clone this library as there is an extended character set. Additional we made another modification to save some more RAM.

> I guess RF interference is always a possibility, ...

My experiences and feedback from other users is that this happens the other way round, means that the MAX7219 chips will disturb mainly the analog input due to the high clocking lines. That was the reason to smooth the analog in.

According the thread in the forum and your statement above you managed to get the 7segment displays working.
So I would propose to wait with merging of this PR until we get other experiences.

peter-masters:
@elral - totally up to you guys whether you want the PR or not, but just for clarity - me saying "I got the segments working" was when I was running my test script, not mobiflight.  In other words, it was when I got the segments reliable.  Following that statement I then tried mobiflight and found the difference between my test script and mobiflight - I thus concluded that it was a software issue from then on, and that resulted eventually in the PR.

DocMoebiuz:
I will close this PR. There are no reports from other users so far. 
Thank you for reporting this and your effort!

PR #236, opened 2023-03-13
Increase multiplexer limit to 99 on the Mega

neilenns:
Now everyone can have all the multiplexers.


neilenns:
Updated board.json file.

[arduino_mega.board.zip](https://github.com/MobiFlight/MobiFlight-FirmwareSource/files/10959429/arduino_mega.board.zip)

elral:
Whow, 99 multiplexer? ;)
That's more as the mega has pins. 
Each multiplexer needs 6 bytes from the device buffer, so this would be OK.
But each multiplexer (16bit) needs ~0.6ms for reading the inputs. I think this will be the limitation not to exceed a certain time. Fast I/O support would speed it up to ~0.2ms which could again increase the max. numbers of multiplexer.

neilenns:
This is just an egreious hack to let a handful of people try more multiplexers. It'll never merge.

PR #238, opened 2023-03-15
Revert mechanism for serial generation for Arduino AVRs

### LINKED ISSUE #237: Revert mechanism for serial generation for Arduino AVRs
In #230 we added a new way for generating Serial numbers for the boards. For the PICO this works fine, because the chips have an individual serial numbers.
For the Arduino family, it doesn't work well... we have a batch of Mega boards that all ended up with the same serial.

We want to:
- use a similar logic as before
- instead of using A0 for random seed, we will be using the "millis()" function which will lead to unique serials
- the meachanism for the raspberry Pico will be untouched

elral:
For AVR's a serial number will be used again instead the UniqueID which was introduced with 2.4.0.
Within a batch all mega's had the same UniqueID.

Fixes #237


DocMoebiuz (approved):
Very cool, very cool!

PR #240, opened 2023-03-30
Fast IO access

### LINKED ISSUE #239: Accelerate reading/writing to I/O's
In preparation for implemeting encoders on multiplexer the access to I/O's should be increased.

The required time for each device was measured and is listed under [Tme consumption](https://github.com/MobiFlight/MobiFlight-Connector/wiki/Time-Consumption). Encoders should be read in every 1ms like implemented in the firmware, but tests showed that this could be increased to 2ms.

According the above mentioned link you can see that with the actual status of the firmware a 4 multiplexer with 16 bit needs 2.4ms to read in all I/O's. To increase the number of available multiplexer e.g. for keymatrixes with more than 4 rows but still to have multiplexer with encoders connected it would be required to fasten up the access to the I/O's.

This could be accomplished by using a direct access to the port registers for "critical" devices instead of using the `digitalRead()/digitalWrite()` function of the Arduino library.

The drawback would be an increased RAM consumption, mainly from the deviceBuffer. Therefore a precondition is to free up RAM by merging PR #219 .

The FastIO will not fit into the flash for the ProMicro (Memory consumption [](https://github.com/MobiFlight/MobiFlight-Connector/wiki/Memory-Consumption) w/o further flash reduction (not checked if possible). On the other hand there it is very likely that not enough flash is available for the required changes to connect encoders to multiplexers.

The Uno/Nano is very limited in RAM, even with eliminating the nameBuffer max. 16(14) encoders could be defined. 6 encoders can be connected directly to the I/O's  (18 pins - 5 pins from MUX device) and the remaining 10(8) can be connected to a multiplexer. So I see no reason to increase the max. number of multiplexer (currently 3). That means the FastIO is not really required but could be beneficial.
*) numbers in brackets are for FastIO

The Mega has enough flash and memory, so the max. number of multiplexer could be increased with implementing the FastIO (required to read in encoders on multiplexers).

The Pico is already much faster and has much more flash and RAM. With the FastIO for the AVR architecure they come nearly to the same access time as the Pico.

elral:
## Description of changes

For each device with multiple accesses to I/O's the arduino functions `digitalread()` and `digitalWrite()`is replaced by direct port access to reduce the time each changed device needs.
This will allow to increase the number of multiplexers while keeping the functionality of encoders and will give the possibility to implement encoders connected to multiplexers.

As this change will exceed the flash for the ProMicro, it will be only implemented for:
* Mega
* Uno/Nano
per compile flag and could be implemented for the Pico (not changed so far).

More description is available in issue #239 

Fixes #239


GioCC:
A few first-look comments:

- Storing two bytes for every "fast" pin (port + mask) wastes a lot of RAM. 
Having these values ready probably only makes sense for driver pins (shift registers, multiplexers, LED drivers), whose number is limited (and could be even more limited if they were shared) and whose usage is much more intensive. It's not as important e.g. for buttons.  
If a better performance than the standard `digitalRead` is still desired, you could have a second overload of the "fast" versions, with the simple pin number as argument, to be used in "regular" cases. 
This overload would incorporate the `digitalPinToBitMask()` and `portOutputRegister(digitalPinToPort())` calls; these are simple macro lookups and therefore still remarkably faster than the standard functions.
It could be evaluated if these "less optimized" functions have sufficient performance to be used also for driver pins.

- I see that in some cases, both the pin number _and_ the pin port+mask are stored; I wonder if it's possible to drop pin number (not examined the code to this extent yet).

- In `digitalWriteFast()`, the read->update->write operation is supposed to be atomic, therefore it's necessary to add the "Save SREG / cli" sequence as used in the default function.

GioCC:
Thanks to you for trying to find even more improvements!

OK, hoping to spare you some time by trying to be clearer:
let's say we add another function in `MFFastIO.h`...

```C++
inline uint8_t digitalReadFast(uint8_t pinNr)
{
    return (*portOutputRegister(digitalPinToPort(pinNr) & digitalPinToBitMask(pinNr)) ? HIGH : LOW;
}
```

this way, e.g. Buttons would continue to use pin numbers without the need to store pin port and pin mask.

About the `detach()` functions: even there, I would imagine that a "fast" version of the `pinMode()` could be introduced to reset the pin to input. I haven't looked at the code yet though.

Switching unused pins back to input is definitely a thing to recommend though.

elral:
There is no real demand for now to implemet this feature. Everything is working fine.
If this changes, this PR can be reopened.

PR #246, opened 2023-06-03
Remove unused temp variable

### LINKED ISSUE #245: temp variable is unused in readConfigLength()
```
src\Config.cpp: In function 'bool readConfigLength()':
src\Config.cpp:75:14: warning: unused variable 'temp' [-Wunused-variable]
     char     temp       = 0;
```

neilenns:
Fixes #245 

Removes the unused `temp` variable. The variable was introduced [here](https://github.com/MobiFlight/MobiFlight-FirmwareSource/commit/e00bc34c23328801ca5866886c33a173ba641eb5).

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -78,7 +78,7 @@ bool readConfigLength()
 
     while (MFeeprom.read_byte(addreeprom++) != 0x00) {
         configLength++;
-        if (addreeprom > length)                                       // abort if EEPROM size will be exceeded
+        if (addreeprom > length)
         {
             cmdMessenger.sendCmd(kStatus, F("Loading config failed")); // text or "-1" like config upload?
             return false;
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -72,14 +72,13 @@ void _activateConfig();
 // reads the EEPROM until NUL terminator and returns the number of characters incl. terminator, starting from given address
 bool readConfigLength()
 {
-    char     temp       = 0;
     uint16_t addreeprom = MEM_OFFSET_CONFIG;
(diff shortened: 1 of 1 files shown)


GioCC (approved):

src/Config.cpp
```diff
@@ -72,14 +72,13 @@ void _activateConfig();
...
 {
-    char     temp       = 0;
     uint16_t addreeprom = MEM_OFFSET_CONFIG;
     uint16_t length     = MFeeprom.get_length();
     configLength        = 0;
 
     while (MFeeprom.read_byte(addreeprom++) != 0x00) {
         configLength++;
-        if (addreeprom > length) // abort if EEPROM size will be exceeded
+        if (addreeprom > length)                                       // abort if EEPROM size will be exceeded
```
DocMoebiuz, line 81: comment alignment. probably due to auto formatting rule.

This comment is also not necessary because it's obvious what is happening.

if `length` was called `eepromLength` it would be even more clear.

In general I propose to put comments above the line instead of behind.

neilenns: Do you want me to remove the comment as part of this PR? I was only trying to remove the temp variable, I didn't bother touching anything else.

DocMoebiuz: we can remove it - thanks

neilenns, line 81:
```suggestion
        if (addreeprom > length)
```

DocMoebiuz (approved):
I was not aware of the habit to have comments on end of line which makes our syntax formatter format this file every where. I don't like the long lines either, I will discuss it on #development

PR #250, opened 2023-06-29
Wrap defines

### LINKED ISSUE #249: Wrap MFBoard.h #defines so they can be specified via environment variables
Right now all the values in MFBoard.h are just straight `#define` without a wrapping `if #ifndef`. This means they can't be specified outside of the build process by environment variables, allowing external changes to the values without having to modify the source code files.

We should wrap all the defines in MFBoard.h with `#ifndef`.

neilenns:
Fixes #249 

## Description of changes

Add `#ifndef` around all of the defines in the MFBoard.h files. Doesn't result in any change in functionality, just enables setting them via environment variables in the future.

### DIFF
--- a/_Boards/Atmel/Board_Nano/MFBoards.h
+++ b/_Boards/Atmel/Board_Nano/MFBoards.h
@@ -74,14 +74,27 @@
 #define STEPPER_SPEED 400 // 300 already worked, 467, too?
 #define STEPPER_ACCEL 800
 
+#ifndef MOBIFLIGHT_TYPE
 #define MOBIFLIGHT_TYPE     "MobiFlight Nano"
+#endif
+#ifndef MOBIFLIGHT_SERIAL
 #define MOBIFLIGHT_SERIAL   "0987654321"
+#endif
+#ifndef MOBIFLIGHT_NAME
 #define MOBIFLIGHT_NAME     "MobiFlight Nano"
+#endif
+#ifndef EEPROM_SIZE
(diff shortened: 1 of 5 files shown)


_Boards/Atmel/Board_Nano/MFBoards.h
```diff
@@ -36,30 +36,65 @@
...
+#ifndef MAX_INPUT_SHIFTERS
 #define MAX_INPUT_SHIFTERS  2
+#endif
+#ifndef MAX_DIGIN_MUX
 #define MAX_DIGIN_MUX       3
+#endif
 
 #define STEPS         64
 #define STEPPER_SPEED 400 // 300 already worked, 467, too?
 #define STEPPER_ACCEL 800
```
neilenns, line 75: I didn't wrap these because I don't think they are used? I need to open a separate issue for it, they only exist in the nano and uno MFBoard.h files.

elral: `define STEPS 64` is indeed not used, but `STEPPER_SPEED` and `STEPPER_ACCEL`  is used in `MFStepper.cpp`.

But we could think about to move it to the `MFStepper.h` file as it is more related to the stepper as to the boards and the speed and acceleration can be set by the connector now.

MobiFlight-Admin (approved):
Thanks!

DocMoebiuz (approved):
Thanks!

PR #253, opened 2023-08-21
Get rid of fixed device numbers

### LINKED ISSUE #174: Removal of the fixed per-device limit of configurable devices (and associated data structures)

elral:
## Description of changes

The number of devices of each type will be calculated before the config gets loaded from the EEPROM.
With this information the device buffer gets filled up with the defined devices.
The device index array isn't needed anymore, instead a pointer to the first device of a type is stored.

fixes #174 

Notes:
* How to handle a full device buffer. For now it can happen that the device buffer gets filled completly, but during loading the devices this is not considered.
* As the device index arrays are not needed anymore, this should be a huge RAM saving. But the compiler does only report a small saving

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -241,8 +241,8 @@ void readConfig()
     if (command == 0)                                        // just to be sure, configLength should also be 0
         return;
 
-    // go through the EEPROM and calculate the number of devices of each type
-    // the call the function to dynamically set up the required arrays of each type
+    // go through the EEPROM and calculate the number of devices for each type
+    // then call the function to dynamically set up the required arrays of each type
     Button::setupArray(5);      // 5 is just an example and must be replaced with the calculated number of devices
 
     do // go through the EEPROM until it is NULL terminated
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -227,6 +227,7 @@ bool readNameFromEEPROM(uint16_t *addreeprom, char *buffer, uint16_t *addrbuffer
     return true;
 }
 
+
 // steps thru the EEPRROM until the delimiter is detected
 // it could be ":" for end of one device config
 // or "." for end of type/pin/config entry for custom device
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -230,10 +230,10 @@ bool getArraysizes()
 {
     if (configLength == 0) // do nothing if no config is available
         return true;
-    uint16_t addreeprom                      = MEM_OFFSET_CONFIG;               // define first memory location where config is saved in EEPROM
-    uint8_t  device                          = readUintFromEEPROM(&addreeprom); // read the first value from EEPROM, it's a device definition
-    bool     copy_success                    = true;                            // will be set to false if copying input names to nameBuffer exceeds array dimensions
-    uint8_t  numberDevices[kTypeStepper + 1] = {0};                             // if new device types are added, change 'kTypeStepper' to the new one!! Otherwise the array inde will be exceeded
+    uint16_t addreeprom              = MEM_OFFSET_CONFIG;               // define first memory location where config is saved in EEPROM
+    uint8_t  device                  = readUintFromEEPROM(&addreeprom); // read the first value from EEPROM, it's a device definition
+    bool     copy_success            = true;                            // will be set to false if copying input names to nameBuffer exceeds array dimensions
+    uint8_t  numberDevices[kTypeMax] = {0};
 
     if (device == 0) // just to be sure, configLength should also be 0
         return true;
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -58,7 +58,7 @@ const uint8_t MEM_OFFSET_CONFIG = MEM_OFFSET_NAME + MEM_LEN_NAME + MEM_LEN_SERIA
 char      serial[3 + UniqueIDsize * 2 + 1] = MOBIFLIGHT_SERIAL; // 3 characters for "SN-", UniqueID as HEX String, terminating NULL
 char      name[MEM_LEN_NAME]               = MOBIFLIGHT_NAME;
 const int MEM_LEN_CONFIG                   = MEMLEN_CONFIG;
-char      nameBuffer[MEM_LEN_CONFIG]       = "";
+char      nameBuffer[MEMLEN_NAMES_BUFFER]  = "";
 uint16_t  configLength                     = 0;
 boolean   configActivated                  = false;
 
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -333,48 +333,49 @@ void readConfig()
             break;
 
 #if MF_SEGMENT_SUPPORT == 1
-        case kTypeLedSegment:
-            params[0] = readUintFromEEPROM(&addreeprom); // Pin Data number
-            params[1] = readUintFromEEPROM(&addreeprom); // Pin CS number
-            params[2] = readUintFromEEPROM(&addreeprom); // Pin CLK number
-            params[3] = readUintFromEEPROM(&addreeprom); // brightness
-            params[4] = readUintFromEEPROM(&addreeprom); // number of modules
-            LedSegment::Add(params[0], params[1], params[2], params[4], params[3]);
+        // this is for backwards compatibility
+        case kTypeLedSegmentDeprecated:
+        // this is the new type
+        case kTypeLedSegmentMulti:
+            params[0] = LedSegment::TYPE_MAX72XX;
+            if (command == kTypeLedSegmentMulti)
+                params[0] = readUintFromEEPROM(&addreeprom); // Type of LedSegment
+
+            params[1] = readUintFromEEPROM(&addreeprom); // Pin Data number
+            params[2] = readUintFromEEPROM(&addreeprom); // Pin CS number
+            params[3] = readUintFromEEPROM(&addreeprom); // Pin CLK number
+            params[4] = readUintFromEEPROM(&addreeprom); // brightness
+            params[5] = readUintFromEEPROM(&addreeprom); // number of modules
+            LedSegment::Add(params[0], params[1], params[2], params[3], params[5], params[4]);
             copy_success = readEndCommandFromEEPROM(&addreeprom); // check EEPROM until end of name
             break;
 #endif
 
 #if MF_STEPPER_SUPPORT == 1
         case kTypeStepperDeprecated1:
-            // this is for backwards compatibility
-            params[0] = readUintFromEEPROM(&addreeprom); // Pin1 number
-            params[1] = readUintFromEEPROM(&addreeprom); // Pin2 number
-            params[2] = readUintFromEEPROM(&addreeprom); // Pin3 number
-            params[3] = readUintFromEEPROM(&addreeprom); // Pin4 number
-            params[4] = readUintFromEEPROM(&addreeprom); // Button number
-            Stepper::Add(params[0], params[1], params[2], params[3], 0);
-            copy_success = readEndCommandFromEEPROM(&addreeprom); // check EEPROM until end of name
-            break;
-
         case kTypeStepperDeprecated2:
-            params[0] = readUintFromEEPROM(&addreeprom); // Pin1 number
-            params[1] = readUintFromEEPROM(&addreeprom); // Pin2 number
-            params[2] = readUintFromEEPROM(&addreeprom); // Pin3 number
-            params[3] = readUintFromEEPROM(&addreeprom); // Pin4 number
-            params[4] = readUintFromEEPROM(&addreeprom); // Button number
-            Stepper::Add(params[0], params[1], params[2], params[3], params[4]);
-            copy_success = readEndCommandFromEEPROM(&addreeprom); // check EEPROM until end of name
-            break;
-
         case kTypeStepper:
+            // Values for all stepper types
             params[0] = readUintFromEEPROM(&addreeprom); // Pin1 number
             params[1] = readUintFromEEPROM(&addreeprom); // Pin2 number
             params[2] = readUintFromEEPROM(&addreeprom); // Pin3 number
             params[3] = readUintFromEEPROM(&addreeprom); // Pin4 number
-            params[4] = readUintFromEEPROM(&addreeprom); // Button number
-            params[5] = readUintFromEEPROM(&addreeprom); // Stepper Mode
-            params[6] = readUintFromEEPROM(&addreeprom); // backlash
-            params[7] = readUintFromEEPROM(&addreeprom); // deactivate output
+
+            // Default values for older types
+            params[4] = (uint8_t)0; // Button number
+            params[5] = (uint8_t)0; // Stepper Mode
+            params[6] = (uint8_t)0; // backlash
+            params[7] = false;      // deactivate output
+
+            if (command == kTypeStepperDeprecated2) {
+                params[4] = readUintFromEEPROM(&addreeprom); // Button number
+            }
+
+            if (command == kTypeStepper) {
+                params[5] = readUintFromEEPROM(&addreeprom); // Stepper Mode
+                params[6] = readUintFromEEPROM(&addreeprom); // backlash
+                params[7] = readUintFromEEPROM(&addreeprom); // deactivate output
+            }
             // there is an additional 9th parameter stored in the config (profileID) which is not needed in the firmware
             // and therefor not read in, it is just skipped like the name with reading until end of command
             Stepper::Add(params[0], params[1], params[2], params[3], params[4], params[5], params[6], params[7]);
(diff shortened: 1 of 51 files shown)


GioCC:
OK, I had to look back at ancient history to dust off some old discussions.
This seems to be a less intrusive solution to tackle the problem that I tried to solve more radically with the abandoned, albeit once working, PR #175 (which also fixed #174, and then some).
That one probably looked too intimidating, while this PR - focusing on the sole storage issue - is indeed attractive in its simplicity.
Unfortunately I'm not able to test it right now because I'm on holiday, but I'm definitely going to have a deeper look at the code.
One point that comes to mind (may be a drawback, or not so much) is that, since the Connector does no longer have the notion of a hard limit to enforce for the number of devices, it won't be able to "stop the user in his tracks" during configuration, but it'll have to wait until the end and see whether the FW reports if the total configuration fits.
Alternatively, the Connector could try and compute the actual space requirement in advance, but that would require an even heavier parametrization than current one (plus, more involved computations which BTW would have to depend on the actual target MCU).

src/Config.cpp
```diff
@@ -223,13 +222,296 @@ bool readEndCommandFromEEPROM(uint16_t *addreeprom)
...
+    if (command == 0) // just to be sure, configLength should also be 0
+        return true;
+
+    // go through the EEPROM and calculate the number of devices for each type
+    do // go through the EEPROM until it is NULL terminated
+    {
+        switch (command) {
+        case kTypeButton:
+            numberButtons++;
+            copy_success = readEndCommandFromEEPROM(&addreeprom); // check EEPROM until end of name
```
GioCC, line 257: This statement occurs identical for each and every "case"; maybe better put just one instance at the end

src/Config.cpp
```diff
@@ -223,13 +222,296 @@ bool readEndCommandFromEEPROM(uint16_t *addreeprom)
...
+    uint8_t  numberInputShifter  = 0;
+    uint8_t  numberDigInMux      = 0;
+
+    if (command == 0) // just to be sure, configLength should also be 0
+        return true;
+
+    // go through the EEPROM and calculate the number of devices for each type
+    do // go through the EEPROM until it is NULL terminated
+    {
+        switch (command) {
```
GioCC, line 254: The "switch" could probably be avoided by placing the counter in an array indexed with the `command` value (as long as the `kTypeXxxxx` values are reasonably not sparse.
The whole case could disappear in something like `numberDevice[command]++;`

elral: The `kTypeXxxxx` values are not perfect for an array due to the deprecated devices, but it makes still sense.

elral:
> One point that comes to mind (may be a drawback, or not so much) is that, since the Connector does no longer have the notion of a hard limit to enforce for the number of devices, it won't be able to "stop the user in his tracks" during configuration, ...

Up to now it is also not perfectly done. We have assumptions what could fit into the device buffer, but never checked all combinations of possible devices if they will fit into the device buffer. Therefore we send the `kStatus` message to the connector if the device buffer gets exceeded.

At least this message must be kept, but I was also thinking about to calculate the remaining space in the device  buffer on the connector side like you mentioned. Each device type has a fixed size which is required in the device buffer, the size of the device buffer could be part of the json file. One exception is the upcoming custom device. In this case the size is unknown which could lead to report the size within the json file (and due to this for all devices?)

I am still searching where the huge memory of all device index arrays is gone to. The proMicro has in sum 70 devices of 11 different types. The saving should be (70-11)\*8=472bytes\*, but the compiler reports only ~70bytes.

Thanks a lot for your comments and a nice vacation!

\*each array entry for the device index arrays needs 8 bytes, for all devices one pointer of this size still required

src/Config.cpp
GioCC: > The `kTypeXxxxx` values are not perfect for an array due to the deprecated devices, but it makes still sense.

Exactly; however, each "useless" element would only cost 1 byte of RAM, while the total flash gain could be more interesting (still to be verified).

GioCC:
Just a vague idea: maybe the connector could send a message when adding (or removing) a device, _during_ configuration (before saving it to the board), to which the board could return the required storage size increment/decrement. (Adding "no device" would return the occupation level for the currently saved configuration). Then the connector could show the updated prospect memory occupation (possibly in %).

> I am still searching where the huge memory of all device index arrays is gone to. The proMicro has in sum 70 devices of 11 different types. The saving should be (70-11)*8=472bytes*, but the compiler reports only ~70bytes.

That is something I'm also curious to investigate. Have you already tried using the memory analysis ("Inspect") function of PlatformIO?

elral:
> That is something I'm also curious to investigate. Have you already tried using the memory analysis ("Inspect") function of PlatformIO?

I tried this, but unfortunately the inspection aborts due to a failure (have to check at home which one). I had this a couple of times also on other projects.

elral:
> Just a vague idea: maybe the connector could send a message when adding (or removing) a device, _during_ configuration (before saving it to the board), to which the board could return the required storage size increment/decrement. (Adding "no device" would return the occupation level for the currently saved configuration). Then the connector could show the updated prospect memory occupation (possibly in %).

This could be a way. Since the same memory for devices is available it should be no problem for existing configurations (not to say it's not required). Maybe a worst case calculation with considering the available pins and the most consuming device might help

elral:
> > I am still searching where the huge memory of all device index arrays is gone to. The proMicro has in sum 70 devices of 11 different types. The saving should be (70-11)_8=472bytes_, but the compiler reports only ~70bytes.
> 
> That is something I'm also curious to investigate. Have you already tried using the memory analysis ("Inspect") function of PlatformIO?

Made again the failure to trust hovering about sizeof(). Sometimes it reports wrong values. A pointer is 2 byte instead of 8byte. So the formula must be (70-11)*2bytes = 118 bytes which should be saved. Compared this to 76bytes saving makes much more sense. 42 bytes less saving for the new functions seems reasonable...

src/config.h
```diff
@@ -23,6 +23,8 @@ enum {
     kTypeMuxDriver,           // 13 Multiplexer selector support (generates select outputs)
     kTypeDigInMux,            // 14 Digital input multiplexer support (example: 74HCT4067, 74HCT4051)
     kTypeStepper              // new stepper type with settings for backlash and deactivate output
+    // if new device types are added, change 'kTypeStepper' for the array numberDevices[] in readConfig() to the new one!!
```
GioCC, line 26: Suggestion: add a last entry like this (comments removed for clarity)
```C++
    // ...
    kTypeDigInMux,
    kTypeStepper,
    kTypeMAXCOUNT
```
and use that value for the array size in `getArraysizes()`. This way no manual update will be required when adding devices.

elral: Again a very good idea!

GioCC:
> Made again the failure to trust hovering about sizeof(). Sometimes it reports wrong values. A pointer is 2 byte instead of 8byte.

Yes, that's an important thing to know.
The indications given by IntelliSense in VSCode are based upon the compiler settings configured in `c_cpp_properties.json` (see `https://code.visualstudio.com/docs/cpp/configure-intellisense-crosscompilation` for reference).
More than one option can be configured; in this case they will appear in the info bar on the bottom right.
_The catch is that, though apparently it should be possible to select e.g. the gcc-avr compiler (see "Compiler path" section), that doesn't work: the only options available are the architectures selectable with "IntelliSense mode", and these are all 32- or 64- bit._
I haven't yet found a way to extend this list (and I believe it isn't possible).

During development, whenever possible, I usually write a section of debug code like this:
```C++
constexpr uint8_t Size_T1 = sizeof(MyClass1);
constexpr uint8_t Size_T2 = sizeof(MyType2);
//...

void report(void)
{
    sprintf(sbuf, "Size of %s: %d bytes\n\r", "MyClass1", Size_T1 );   Serial.println(sbuf);
    sprintf(sbuf, "Size of %s: %d bytes\n\r", "MyType2", Size_T2 );   Serial.println(sbuf);
    //...
}
```
so as to have the actual running code report the relevant sizes once and for all.

elral:
According the below calculations for the device buffer w/o fixed device numbers the limitations are coming from the max. pins which are availabe (except for the Nano/Uno) if the device buffer will be increased for the Mega and ProMicro. For the Mega it's no problem, he has plenty of available RAM. For the ProMicro we could benefit from the RAM saving (with spending additional 3 bytes).
The Nano/Uno is some kind of problematic. Increasing the max. number of analog inputs from 6 to 8 via PR 248 wasn't a really good idea. I would roll back again to max. 6 analog inputs which saves 58 bytes. Additionally I would limit the max. numbers of servos to 6 (for now it is 2) which would save 92 bytes. For the missing 10 bytes we can profite from the RAM saving.

This would mean that we stay with the definition of max. devices within each board.json file. It is also still sensefull to limit the max. input (and output?) devices per type to considere the processing time. In this case an online check of the RAM usage of the device buffer is not required, either on the connector side nor on the FW side.

Or we are just using the existing message, that the device buffer is filled and no more device could be added. This will only happpen for the Nano/Uno. Here also an online check of the RAM usage of the device buffer is not required, either on the connector side nor on the FW side. "Problems" with the processing time due to too much input (output?) devices would be in the responsibility of the user. Hints who many devices per type should not be exceeded should be done. In this case no limitation in the board.json files are required.

However, more than these changes would not be required as no communication with the connector is required (except the existing warning which might be required for the Nano/Uno).

************************************************************************

Required device memory for AVR's

|                   |  AVR's     |  Pico     |
| ----------------- |  --------- |-----------|
| analog device     |  29 byte   |  40 byte  |
| button            |  4 byte    |   8 byte  |
| digital input MUX |  6 byte    |   8 byte  |
| encoder           |  31 byte   |  36 byte  |
| input shifter     |  11 byte   |  16 byte  |
| lcdDisplay        |  18 byte   |  24 byte  |
| output            |  2 byte    |   2 byte  |
| Output Shifter    |  9 byte    |   9 byte  |
| Led Segment       |  25 byte   |  36 byte  |
| Servos            |  23 byte   |  64 byte  |
| Stepper           |  82 byte   | 100 byte  |


Required bytes per pin per device

|                   | pins | byte/pin  AVR's| byte/pin Pico |
| ----------------- | ---- | ---------------|---------------|
| analog device     | 1    | 29 byte        |  40 byte      |
| Servos            | 1    | 23 byte        |   64 byte     |
| Stepper           | 4    | 20.5 byte      |   25 byte     |
| encoder           | 2    | 15.5 byte      |   18 byte     |
| lcdDisplay        | 2*   | 9  byte        |   12 byte     |
| Led Segment       | 3    | 8.3 byte       |   12 byte     |
| digital input MUX | 5    | 6 byte         |  1.2 byte     |
| button            | 1    | 4 byte         |   4 byte      |
| input shifter     | 3    | 3.6 byte       |  5.3 byte     |
| Output Shifter    | 3    | 3 byte         |    3 byte     |
| output            | 1    | 2 byte         |    2 byte     |

* for multiple LCD's only 2 pins are required

************************************************************************
**Mega:**
Size of device buffer is: 1500 bytes
w/o fixed device numbers RAM saving is 289 bytes
number of pins: 68
number of analog pins: 16

required bytes in device buffer actually:

- 16 * analog  ->  464 bytes
- 10 * servo   ->  230 bytes
- 10 * stepper ->  820 bytes
-  1 * encoder ->   31 bytes  
max bytes    ->  1545 bytes   -> exceding device buffer by 45 bytes

required bytes in device buffer w/o fixed device numbers:

- 16 * analog  ->  464 bytes
- 48 * servo   -> 1104 bytes
-  1 * stepper ->   21 bytes
max bytes    -> 1589 bytes   -> exceding device buffer, can be increased w/o any impact

************************************************************************
**Nano/Uno:**
Size of device buffer is: 300 bytes
w/o fixed device numbers RAM saving is 55 bytes
number of pins: 18
number of analog pins: 8

required bytes in device buffer actually:

- 8 * analog  ->  232 bytes
- 2 * servo   ->   46 bytes
- 2 * stepper ->  164 bytes
max bytes    -> 442 bytes   -> exceding device buffer by 142 bytes!!

required bytes in device buffer w/o fixed device numbers:

- 8 * analog  ->  232 bytes
- 10 * servo   -> 230 bytes
max bytes    -> 460 bytes   -> exceding device buffer by 160 bytes!!

************************************************************************
**ProMicro:**
Size of device buffer is: 400 bytes
w/o fixed device numbers RAM saving is 65 bytes
number of pins: 18
number of analog pins: 9

required bytes in device buffer actually:

- 9 * analog  ->  261 bytes
- 3 * servo   ->   69 bytes
- 1 * stepper ->   82 bytes
- 1 * encoder ->   31 bytes
max bytes    -> 443 bytes   -> exceding device buffer by 43 bytes!!

required bytes in device buffer w/o fixed device numbers:

- 9 * analog  ->  261 bytes
- 9 * servo   ->  207 bytes
max bytes    -> 468 bytes   -> exceding device buffer by 68 bytes!!

************************************************************************
**Raspberry Pico:**
Size of device buffer is: 2000 bytes
w/o fixed device numbers RAM saving is 324 bytes
number of pins: 26
number of analog pins: 3

required bytes in device buffer actually, does not change for w&o fixed device numbers:

- 3 * analog  ->  120 bytes
- 8 * servo   ->  512 bytes
- 3 * stepper ->  300 bytes
- 1 * encoder ->   36 bytes
- 1 * button  ->    8 bytes
max bytes    -> 976 bytes   -> device buffer could be reduced.


************************************************************************

Limitations for max. devices are
* max. available pins
* max. available analog pins
* processing time for inputs (each loop these are processed) -> most important
  * encoder
  * input shifter
  * input multiplexer
* processing time for outputs (only for new values processed)
  * LCD's
  * output shifter
  * 7 segment displays
  * stepper

|               | Atmega |  Pico           | Remark                                                        |
| ------------- | -------|  ---------------|  ------------------------------------------------------------ |
| Outputs       | 11 us  |                 |  only writing one output, analogWrite(), no changes           |
| Buttons       | 9 us   |  2 us           |  only reading one input                                       |
|               | 224 us |  220 us (??)    |  reading AND evaluating one input, must be wrong for the Pico |
| LEDSegment    | 6.3 ms |  1.3 ms         |   2x8 daisy chained                                           |
| Encoder       | 23 us  |    3 us         |  only reading one encoder                                     |
|               | 290 us |   22 us         |   reading AND evaluating one                                  |
| Stepper       | 23 ms  |    3 us         |  not changed to FastIO                                        |
| Servos        |        |                 |                                                               |
| LCD           | 12 ms  |  9.7 ms         |  limited by I2C                                               |
| AnalogIn      | 122 us |   11 us         |  calculating average value                                    |
| OutputShifter | 0.74 ms|                 |  4x8bit daisy chained                                         |
| InputShifter  | 0.5 ms |                 |  4x8bit daisy chained                                         |
| Multiplexer   | 2.4 ms |                 |  4 x 16 bit                                                   |

elral:
* First I double checked the required bytes in the device buffer for each device (1st table).
* Second I calculated the required bytes per pin to see which device has the biggest memory consumption (2nd table).
* Third I checked if there are limitations by the used libs:
   * servo lib is limited to 48 servos for the mega
   * servo lib is limited to 12 servos for the other AVR's
   * servo lib is limited to 8 servos for the Pico (as the state machines are used)
* Fourth I checked the limitation by available pins (analog pins)

The analog device required the most bytes in the memory, so I used the max analog pins from each device to calculate the required bytes.
The second highest memory consumption have the servos, but they are limited by the lib and by the reaming pins except the mega and pico. This limitation was used to calculate the required bytes.
Next I calculated the remaining pins to see how many steppers, which are the devices with third highest memory consumption, can be defined.
At least I calculated how many encoders and buttons can be defined due to the remaining pins.

**_All these steps would result in the required size of the device buffer without a limitation except from the libs or available pins. So the user can use all pins with whatever device he wants._**

**And yes, thinking about again it is not 100% correct. The limitation from the libs makes it a little bit difficult.**
With less analog devices the user could define e.g. more servos for the Nano/Uno and ProMicro.
This could be for the Nano/Uno:

- 6 * analog  ->  174 bytes
- 12 * servo   -> 276 bytes
- max bytes   -> 450 bytes which is less than 8 * analog and 10 * servo.

Or 4 steppers (16 pins) and 2 analog devices is 4 * 82 = 328 plus 2 * 29 = 58 byte in sum 386 byte which is also less then the above calculation.
So I am pretty confidenet that the above calculated buffer sizes are the max. sizes which are required.

----------------------------------------------------
Summary for the size of the device buffer

|               | actual |  saving     | additionally*| sum       | required  | remark    |
| ------------- | -------|  -----------| -----------  |  -------- |  -------- |  -------- |
| Mega          | 1500   |  289        |    496       |  2285     |  1589     |  overall memory consumption can be reduced |
| Nano/Uno      |  300   |   55        |     66       |   421     |   460     |  max. servo must be limited          |
| ProMicro      |  400   |   65        |    110       |   575     |   468     |  overall memory consumption can be reduced |
| Pico          | 2000   |  324        |    496       |  2820     |   976     |  overall memory consumption can be reduced |
 
 *additionally means: The size of the nameBuffer[] is defined by MEM_LEN_CONFIG, but it was intended to use MEMLEN_NAMES_BUFFER. But this is already used to check for exceeding the array, so negative impact has to be expected
 to change the array size as original intended.

Except for the Nano/Uno the max. device numbers in the board.json file can be increased to (available pins/required pins). For the Nano/Uno max. servos would be also 8.

elral:
Oh, I forgot the scan time for the multiplexer.
Yes, I also wondered that they are so high. But I see a further optimazation in a separate PR ;)

DocMoebiuz (approved):

PR #256, opened 2023-08-25
Wrong define for debug message to connector

### LINKED ISSUE #254: Wrong define for debug message to connector
There is a wrong `#define` within DigInMux.cpp for printing debug messages to the connector.

It must be `DEBUG2CMDMESSENGER `instead of `DEBUG2MSG`.

Additional it would make sense to print the buffer usage per device if `DEBUG2CMDMESSENGER` is activated.
An additional debug message in `allocateMem()` would be required.

elral:
## Description of changes

Define `DEBUG2MSG` replaced by `DEBUG2CMDMESSENGER `to get this info during debuging to the connector.

Added an additionally print to the connector to get the size of devices which gets registered in the device buffer.

Fixes #254

### DIFF
--- a/src/MF_DigInMux/DigInMux.cpp
+++ b/src/MF_DigInMux/DigInMux.cpp
@@ -40,7 +40,7 @@ namespace DigInMux
         MFDigInMux::attachHandler(handlerOnDigInMux);
         digInMuxRegistered++;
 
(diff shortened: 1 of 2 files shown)


GioCC (approved):

DocMoebiuz (approved):
Thanks

PR #257, opened 2023-08-25
Fixed exceeding digInMux array

### LINKED ISSUE #255: Clearing DigInMux can lead to a crash
If a new config gets uploaded, all deveces will be cleared before.

Clearing the defined input multiplexer is done within `DigInMux::clear()` where all defined input multiplexer gets detached. In the for loop the array `digInMux[]` gets exceeded as `digInMuxRegistered` for the array index is used and not `i`

elral:
## Description of changes

In the `for` loop the array `digInMux[]` gets not exceeded anymore as `i` is used instead of`digInMuxRegistered` for the array index.

Fixes #255


DocMoebiuz (approved):
Wow! What an error :D

PR #260, opened 2023-10-19
Flash optimizations

elral:
## Description of changes

I went a couple of times through all functions, not much I found to optimize the Flash usage.
* there are unused parameters in the handler functions. They are deleted -> -30 bytes
* common usage of size for all pins (int -> uint8_t) -> no reduction of flash
* not using UniqueID lib for AVR's -> -192 bytes
* replacing the sprintf() function for the serial handling in config.cpp by an own simple one -> - 1468 bytes

Overall this results in 1690 bytes less of Flash usage and 43 bytes less of RAM usage.

**One important remark**
I have deleted the usage of UniqueID for all AVR Boards. There was only less time where this was used. If a board was initialized in this timeframe, a new serial number will be generated with this change. I guess there are only verly less user who are affected, not to say no user is affected.
I vote to take now the chance to get rid of these serial numbers to not deal with them in the future.
If we should still keep them, I will add again the detection but without using the sprintf() function to take care of the huge flash saving.

@MobiFlight-Admin @neilenns @GioCC may I ask you to comment on this and to add ideas/changes for further reducing of Flash/RAM usage.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -464,7 +464,7 @@ void generateRandomSerial()
     serial[1]             = 'N';
     serial[2]             = '-';
     serial[6]             = '-';
-    serial[10] = 0x00;
+    serial[10]            = 0x00;
     uint16_t randomSerial = random(4095);
     for (uint8_t i = 3; i < 6; i++) {
         serial[i] = (randomSerial & 0x000F) + 48; // convert from 4bit to HEX string
@@ -474,7 +474,7 @@ void generateRandomSerial()
     randomSerial = random(4095);
     for (uint8_t i = 7; i < 10; i++) {
         serial[i] = (randomSerial & 0x000F) + 48; // convert from 4bit to HEX string
-        if (serial[i] >= 58) serial[i] += 8;      // if HeX value is A - F add 8 to get the letters
+        if (serial[i] >= 58) serial[i] += 7;      // if HeX value is A - F add 7 to get the letters
         randomSerial >>= 4;
     }
     MFeeprom.write_block(MEM_OFFSET_SERIAL, serial, MEM_LEN_SERIAL);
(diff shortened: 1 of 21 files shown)


src/Config.cpp
```diff
@@ -451,17 +457,40 @@ bool getStatusConfig()
 void generateRandomSerial()
 {
     randomSeed(millis());
-    sprintf(serial, "SN-%03x-%03x", (unsigned int)random(4095), (unsigned int)random(4095));
+    serial[0]             = 'S';
```
DocMoebiuz, line 460: i don't like it - but freeing up more than 1.2k is a compelling reason to start liking it.

src/MF_Button/MFButton.h
```diff
@@ -10,7 +10,7 @@
 
 extern "C" {
 // callback functions always follow the signature: void cmd(void);
-typedef void (*buttonEvent)(byte, uint8_t, const char *);
+typedef void (*buttonEvent)(byte, const char *);
```
DocMoebiuz, line 13: i think we should replace `byte` with `uint8_t` everywhere in the code.

src/MF_Segment/MFSegments.cpp
```diff
@@ -11,7 +11,7 @@ MFSegments::MFSegments()
     _moduleCount = 0;
 }
 
-void MFSegments::display(byte module, char *string, byte points, byte mask, bool convertPoints)
+void MFSegments::display(uint8_t module, char *string, uint8_t points, uint8_t mask, bool convertPoints)
```
DocMoebiuz, line 14: Ha, you have done it here and in the following changes.

DocMoebiuz (approved):
Looking good. Thanks!

DocMoebiuz (changes requested):
There is one `byte` that we want to change to `uint8_t` so that it is perfect.

DocMoebiuz (approved):
THANK YOU!

PR #261, opened 2023-10-20
TM1637 memory and code structure improvements

### LINKED ISSUE #251: Support for displays based on the TM1637 driver

DocMoebiuz:
## Description of changes

Fixes #251 

_Describe changes here_

- Introduced type instead of using the CS pin
- Memory improvements.

### Before (pro micro)
```
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [========  ]  77.3% (used 1978 bytes from 2560 bytes)
Flash: [==========]  99.9% (used 28630 bytes from 28672 bytes)
```

### After
```
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [========  ]  77.3% (used 1978 bytes from 2560 bytes)
Flash: [==========]  99.1% (used 28422 bytes from 28672 bytes)
```

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -78,8 +78,7 @@ bool readConfigLength()
 
     while (MFeeprom.read_byte(addreeprom++) != 0x00) {
         configLength++;
-        if (addreeprom > length)
-        {
+        if (addreeprom > length) {
             cmdMessenger.sendCmd(kStatus, F("Loading config failed")); // text or "-1" like config upload?
             return false;
         }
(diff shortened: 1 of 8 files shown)


elral (approved):
No more comments from my side.

elral (approved):
OK

PR #264, opened 2023-10-25
Fix board files

### LINKED ISSUE #263: Within boad.h files wrong defines for Custom Devices
In all board.h files the Custom Devices are enabled by `#define MF_CUSTOMDEVICE_SUPPORT 2`  except for the mega where it is correctly `#define MF_CUSTOMDEVICE_SUPPORT 1.`

In some parts of the code we check for `#if MF_CUSTOMDEVICE_SUPPORT == 1` which will not be true as `MF_CUSTOMDEVICE_SUPPORT` is 2.

For now it doesn't matter as in the platformio.ini `MF_CUSTOMDEVICE_SUPPORT` is set to `0` or `1`.

Furthermore a CustomDevice is not a basic device, so if it is not defined it should still be undefined. That means that:

```
#ifndef MF_CUSTOMDEVICE_SUPPORT
#define MF_CUSTOMDEVICE_SUPPORT 1 or 2
#endif
```

at least must not be in the board.h files.

elral:
## Description of changes

Custom Devices are not a basic device, so `MF_CUSTOMDEVICE_SUPPORT` should be set to 0 or should be just not defined. So `#define MF_CUSTOMDEVICE_SUPPORT 1` / `#define MF_CUSTOMDEVICE_SUPPORT 2`  is deleted in all board.h files and in the platformio.ini file.

In the xyz_custom.ini files this definition is already included to get the custom devices compiled.

/edit: Fixes #263

### DIFF
--- a/_Boards/Atmel/Board_Mega/MFBoards.h
+++ b/_Boards/Atmel/Board_Mega/MFBoards.h
@@ -82,9 +82,6 @@
 #ifndef MOBIFLIGHT_NAME
 #define MOBIFLIGHT_NAME     "MobiFlight Mega"
 #endif
-#ifndef EEPROM_SIZE
-#define EEPROM_SIZE         4096 // EEPROMSizeMega
-#endif
 #ifndef MEMLEN_CONFIG
 #define MEMLEN_CONFIG       1496 // max. size for config which wil be stored in EEPROM
 #endif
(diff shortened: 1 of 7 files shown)


DocMoebiuz (approved):
Thanks!

PR #272, opened 2023-11-08
Beta - Fix memory overlap in DeviceBuffer

### LINKED ISSUE #271: With FW 2.4.3 LCD's might not work or Mega can crash
With the latest FW version 2.4.3 it was reported on Discord that:
1) LCD's are getting initialized but afterwards they are frozen or showing ugly content
2) After booting up the mega continously messages are send to the serial connection

elral:
## Description of changes

With PR #253 the limitation for the max devices per type was eliminated.
Before this PR memory was allocated in the device buffer for each single device. The highest amount of memory is 31 byte for one encoder.
Now the memory gets allocated in the device buffer for all devices of one type at once. If the required memory is more than 255 bytes, e.g. 9 encoders need 9 * 31bytes = 279 bytes, only 23 bytes gets reserved as the size to allocate memory is limited to a byte.

In `allocateMem.cpp/.h` the variable `size` within the functions `allocateMemory()` and `FitInMemory()` must be defined as `uint16_t`.

Fixes #271

### DIFF
--- a/src/allocateMem.cpp
+++ b/src/allocateMem.cpp
@@ -15,9 +15,9 @@ std::size_t    deviceBuffer[MF_MAX_DEVICEMEM] = {0};
 uint16_t nextPointer                    = 0;
 
 #if defined (ARDUINO_ARCH_AVR)
-uint8_t     *allocateMemory(uint8_t size)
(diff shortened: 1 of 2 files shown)


DocMoebiuz (approved):
Nice!

PR #274, opened 2023-11-10
Support keep alive messages from the desktop app

### LINKED ISSUE #273: Buttons, encoders, and diginmux actions do not prevent power save mode
The button, encoder, and digmuxin code are missing calls to `setLastCommandMillis()`, which means interacting with any of those devices isn't sufficient to prevent power save mode from kicking in.

We should probably call `setLastCommandMillis()` for those input devices too.

Reported by @elral in discord.

neilenns:
Fixes #273

This PR fundamentally changes how the firmware knows when to keep displays awake.

Instead of relying on the firmware to track when various events happen (and missing some events, like buttons, encoders, and multiplexers), it simply relies on the desktop app to tell it to stay awake.

The companion desktop app change sends a keep alive message every 5 minutes. If it's been 15 minutes since one of those messages came in the firmware will put displays to sleep.

Bonus: Removing all the calls to `setLastCommandMillis` saves about 1k of space (at least for the pico)

### DIFF
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -104,7 +104,7 @@ void OnUnknownCommand()
     cmdMessenger.sendCmd(kStatus, F("n/a"));
 }
 
-// Handles requests from the desktop app to disable powersaving mode
+// Handles requests from the desktop app to disable power saving mode
 void OnSetPowerSavingMode()
 {
     bool enablePowerSavingMode = cmdMessenger.readBoolArg();
@@ -114,7 +114,7 @@ void OnSetPowerSavingMode()
     // power saving mode to get turned off.
     if (!enablePowerSavingMode) {
 #ifdef DEBUG2CMDMESSENGER
-        cmdMessenger.sendCmd(kDebug, F("Disabling power save mode via request"));
+        cmdMessenger.sendCmd(kDebug, F("Disabling power saving mode via request"));
 #endif
         setLastCommandMillis();
     }
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -109,27 +109,29 @@ void OnSetPowerSavingMode()
 {
     bool enablePowerSavingMode = cmdMessenger.readBoolArg();
 
+    // If the request is to enable powersaving mode then set the last command time
+    // to the earliest possible time. The next time loop() is called in mobiflight.cpp
+    // this will cause power saving mode to get turned on.
+    if (enablePowerSavingMode) {
+#ifdef DEBUG2CMDMESSENGER
+        cmdMessenger.sendCmd(kDebug, F("Enabling power saving mode via request"));
+#endif
+        lastCommand = 0;
+    }
     // If the request is to disable power saving mode then simply set the last command
     // to now. The next time loop() is called in mobiflight.cpp this will cause
     // power saving mode to get turned off.
-    if (!enablePowerSavingMode) {
+    else {
 #ifdef DEBUG2CMDMESSENGER
         cmdMessenger.sendCmd(kDebug, F("Disabling power saving mode via request"));
 #endif
-        setLastCommandMillis();
+        lastCommand = millis();
     }
-
-    // There is no handling of a request to enable powersaving mode right now.
 }
 
 uint32_t getLastCommandMillis()
 {
     return lastCommand;
-}
-
-void setLastCommandMillis()
-{
-    lastCommand = millis();
 }
 
 void OnTrigger()
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -77,6 +77,7 @@ void attachCommandCallbacks()
     cmdMessenger.attach(kSetName, OnSetName);
     cmdMessenger.attach(kGenNewSerial, OnGenNewSerial);
     cmdMessenger.attach(kTrigger, OnTrigger);
+    cmdMessenger.attach(kSetPowerSavingMode, OnSetPowerSavingMode);
 
 #if MF_LCD_SUPPORT == 1
     cmdMessenger.attach(kSetLcdDisplayI2C, LCDDisplay::OnSet);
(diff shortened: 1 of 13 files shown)


elral:
I still vote that the connector actively sets the power mode, e.g. by:
`void OnSetPowerSavingMode() { SetPowerSavingMode(cmdMessenger.readBoolArg());}`

The connector knows when to enter the power safe mode in 15min. by not sending keep alive messages anymore. So the connector can also the time when to send the message to enter the power sage mode. This time is for now 15min which is comiled into the firmware, now it could be an user choosable time from the settings dialog (or just still the fixed 15min.).
To leave the power saving mode the connector sends the required message instead of starting keep alive messages.
This would also reduce the serial traffic.

If it would be done according this, there would be the follwoing changes:
* CommandMessenger.cpp
** The function like above
** `getLastCommandMillis()` and `void setLastCommandMillis()`would not be needed anymore
* Mobiflight.cpp
** `void updatePowerSaving()` would not be needed anymore
** `updatePowerSaving();` in the main loopwould not be needed anymore
** `SetPowerSavingMode(false);` to be added in `ResetBoard()`
* Mobiflight.h
** `void SetPowerSavingMode(bool state);`to be added

neilenns:
I thought about changing to have the desktop manage everything, including putting things to sleep, but decided against it for a couple of reasons:

1. Backwards compatibility is challenging (as you mentioned in your edit)
2. If the connector ever crashes for some reason (which sadly does happen) the boards will never get a go to sleep message and will stay on permanently

There is very little serial traffic with the current approach, it's just one message every 5 minutes. We could even safely extend it to 10 or even 14 minutes if we wanted to I think.

DocMoebiuz:
I originally implemented the power saving mode to make sure that if anything stays on, it will turned off. 

This has to work without Mobiflight Connector running.

The Connector should prevent Power Saving Mode while in Run Mode. 

If the connector crashes and if it could not send a proper stop command - the boards should turn all outputs off - they have to be able to do it without the connector.

I don't see a reason why the Connector should ever send a PowerSavingMode ON message. The Connector would be in STOPped mode for 15 minutes and then send a PowerSavingMode ON. But it is in STOP - so no timers are really running etc

elral:
> 2. If the connector ever crashes for some reason (which sadly does happen) the boards will never get a go to sleep message and will stay on permanently

I thought also about this. I think after a crashed connector the most user will restart the connector. Yeah, but not always. So indeed better to have the keep alive message.

> There is very little serial traffic with the current approach, it's just one message every 5 minutes. We could even safely extend it to 10 or even 14 minutes if we wanted to I think.

That's true, compared to other devices it's nearly nothing.

What we can think about to reduce the time in the FW to 1 minute and sending every e.g. 30sec. Then the user could define the time when the boards should go sleep in the settings, with a minum of one minute and a tolerance of 30 sec.
But maybe again over engineered ;)

DocMoebiuz:
Why can't we just make it work as it was intended? That would be the best. This feature was never meant to be really "visible" to the user.

Being able to set the timeout period - what's the use case?

At the moment it's not working and people are noticing the feature because of its bug. Once we fix the bug and it behaves as intended no one will care anymore about it.

In my opinion there is no need to make this configurable.

Send a keep alive message every 5 mins. Reset the timeout in the firmware. Done.

What am I missing?

elral:
I am also fine with this approach

neilenns:
> Send a keep alive message every 5 mins. Reset the timeout in the firmware. Done.

That's what I did 😀

src/CommandMessenger.cpp
```diff
@@ -102,6 +104,24 @@ void OnUnknownCommand()
...
+    // to now. The next time loop() is called in mobiflight.cpp this will cause
+    // power saving mode to get turned off.
+    if (!enablePowerSavingMode) {
+#ifdef DEBUG2CMDMESSENGER
+        cmdMessenger.sendCmd(kDebug, F("Disabling power saving mode via request"));
+#endif
+        setLastCommandMillis();
+    }
+
+    // There is no handling of a request to enable powersaving mode right now.
```
DocMoebiuz, line 122: Why? This method makes one assume, you could set the mode and enter or leave PowerSavingMode.
i think we should name it so that it is clear what it does, e.g. `OnPreventPowerSavingMode()`

src/Config.cpp
```diff
@@ -112,7 +112,6 @@ void OnSetConfig()
 #ifdef DEBUG2CMDMESSENGER
     cmdMessenger.sendCmd(kDebug, F("Setting config start"));
 #endif
-    setLastCommandMillis();
```
DocMoebiuz, line 115: I like that removing this makes it simple, and saves money.

src/MF_Encoder/MFEncoder.cpp
```diff
@@ -15,7 +15,7 @@
 // -----
 // 18.01.2014 created by Matthias Hertel
 // -----
-
+#include "mobiflight.h"
```
DocMoebiuz, line 18: Was this by accident?

DocMoebiuz (changes requested):
See my comments please.

src/MF_Encoder/MFEncoder.cpp
neilenns: DANG IT

src/CommandMessenger.cpp
neilenns: Since I'm reusing the existing `kSetPowerSavingMode` event I named the handler `OnSetPowerSavingMode` to stay consistent with how all the other events are named.

It doesn't support enabling power saving mode because that would be a bunch more code changes to the firmware I didn't really want to deal with. I guess I could make it work though by setting the last command time to the earliest possible time, then that would automagically turn power saving mode on next time loop() was called.

What's your preference? Add that? Or rename the method and live with it not being consistently named like the rest?

neilenns: Turns out it was really easy to support the request to go to sleep, so I added that. Now the method name makes sense :)

DocMoebiuz (approved):

PR #276, opened 2023-11-12
Beta - Board wont crash if limit switch for steppers is configured

### LINKED ISSUE #275: Beta - Board crashes if limit switch for steppers is configured
If a limit switch for the stepper device is configured and gets uploaded, the board crashes. This is valid for all boards and for FW version 2.4.2 and higher.

elral:
## Description of changes

With the code changes done with PR #261 the structure for reading the stepper config was bundled and slightly changed.
While the config gets read from the eeprom, the parameter are assigned to an array and this array is used for the parameters of the `Stepper::Add(params[0], params[1], params[2], params[3], params[4], params[5], params[6], params[7]);` function.
For the actual stepper type the button gets not read which is now added with this PR. Furthermore sending commands to an not initialized stepper will be blocked

If the button gets not read, the next read command, which is reading the stepper mode, reads the button number. So the button number is treated as stepper mode. Within the stepper code it's checked if it's a valid mode. If not, and that's for all button numbers bigger then 2, creating a new stepper class is aborted. This is saved but not considered when commands are issued. 

Fixes #275

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -387,7 +387,7 @@ void readConfig()
             params[6] = (uint8_t)0; // backlash
             params[7] = false;      // deactivate output
 
-            if (command == kTypeStepperDeprecated2) {
+            if (command == kTypeStepperDeprecated2 || command == kTypeStepper) {
                 params[4] = readUintFromEEPROM(&addreeprom); // Button number
             }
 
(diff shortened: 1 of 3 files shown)


src/Config.cpp
```diff
@@ -387,7 +387,7 @@ void readConfig()
             params[6] = (uint8_t)0; // backlash
             params[7] = false;      // deactivate output
 
-            if (command == kTypeStepperDeprecated2) {
+            if (command == kTypeStepperDeprecated2 || command == kTypeStepper) {
```
DocMoebiuz, line 390: Oops!

DocMoebiuz (approved):

Issue #284, opened 2023-11-22
Pin 1 gets set as output or input when a button/analogIn is defined


elral:
With PR #253 the fixed device numbers were removed. All devices from one type are now behind each other in the deviceBuffer to reserve sufficient memory and only the pointer for all devices is saved.

To get the pointer it is called: `outputs = new (allocateMemory(sizeof(MFOutput) * count)) MFOutput;`

So the constructor gets called and executed. As the definition of the constructor is `MFOutput(uint8_t pin = 1);`  and the code is:

```
MFOutput::MFOutput(uint8_t pin)
{
    _pin   = pin;
    _value = false;
    pinMode(_pin, OUTPUT); // set pin to input
    set(_value);
}
```

Pin 1 gets set as Output if calculating the required size in the devicebuffer.

The same applies for the AnlogIn.

neilenns:
Woah, subtle bug!

elral:
I reduced everything to the minimum:

```
#include <Arduino.h>
#include "Output.h"

void setup() {
  Serial.begin(115200);
  while(!Serial) delay(10);
  delay(1000);
  Serial.println("Serial initialized");
  Output::setupArray(4);
  Output::Add(2);
  Output::Add(3);
  Output::Add(18);
  Output::Add(19);
}

void loop() {
  Serial.println("Setting output to 255");
  Output::OnSet(2,255);
  Output::OnSet(3,255);
  delay(1000);
  Serial.println("Setting output to 0");
  Output::OnSet(2,0);
  Output::OnSet(3,0);
  delay(1000);
}
```

with slight adaptions to get it run without the CMDMessenger and the onSet() function.

And it still happens.
The good things are, I have a base to to debug it better or to ask earlephilhower, and it is NOT related to our code.

PR #285, opened 2023-11-22
Two pins get set on Pico Board

### LINKED ISSUE #283: Two Pins get set at the same time on Pico
An user reported on [Discord](https://discord.com/channels/608690978081210392/1175413489674756137) that 2 pins are set on the same time on the Pico Board. 
The pin numbers have a difference of 16 and both pins must be defined as output. E.g. if pin 2 and 18 are definded as output, setting pin 2 will set also pin 18. Same applies to e.g. pin 3 and 19.

Tests have shown that this happens at least with last stable FW version 2.4.1.
Since FW version 2.4.0 the Pico is a supported device.

elral:
## Description of changes

On the Pico two pins get set at the same time if both pins are configured as output and the difference of the pin numbers are 16.

Moving the initialization of the pin from the class `MFOutput()` to the function which calls the initialization `Output::Add()` solves this issue.
For now it is unclear why this change fixes it. More investigation has to be done.
The change is only done for the Pico and not for the other boards.

Additionally changes are implemented to fix issue #284 to avoid merge errors when both PR gets merged. An additional comment will be given in the upcoming PR.

Fixes #283

### DIFF
--- a/src/MF_Output/MFOutput.cpp
+++ b/src/MF_Output/MFOutput.cpp
@@ -6,14 +6,18 @@
 
 #include "MFOutput.h"
 
-MFOutput::MFOutput() {}
+MFOutput::MFOutput()
+{
+    _value = false;
+}
 
-void MFOutput::init(uint8_t pin)
+void MFOutput::attach(uint8_t pin)
 {
     _pin   = pin;
-    _value = false;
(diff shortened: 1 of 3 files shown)


DocMoebiuz (approved):
Nice!

PR #286, opened 2023-11-22
init devices with attach() function and not via constructor

### LINKED ISSUE #284: Pin 1 gets set as output or input when a button/analogIn is defined

elral:
## Description of changes

For the devices `AnalogIn` and `Buttons` an `attach()` function ia added where all required init paramaters are transferred. These init parameters were part of the constructor.
Now calling the constructor will only set that the device is not initialized.

The calling of the constructor is since PR #253 also done to calculate the required memory for all devices of one type and to reserve it.  As no init parameters are available at this point the standard paramaters were taken and e.g. pin 1 for the output and analogIn device was set output / input.

The required changes for outputs are part of PR #285.

For the InputShifter the default value in the constructor is deleted. Not required here as it gets set also in the attach() function.

For the servo device the not required constructor was deleted.

For the stepper the default parameters in the attach() function are deleted to be the same as the other devices.

Fixes #284

### DIFF
--- a/src/MF_Analog/Analog.cpp
+++ b/src/MF_Analog/Analog.cpp
@@ -37,7 +37,8 @@ namespace Analog
         if (analogRegistered == maxAnalogIn)
             return;
 
-        analog[analogRegistered] = MFAnalog(pin, name, sensitivity);
+        analog[analogRegistered] = MFAnalog();
+        analog[analogRegistered].attach(pin, name, sensitivity);
         MFAnalog::attachHandler(handlerOnAnalogChange);
         analogRegistered++;
 #ifdef DEBUG2CMDMESSENGER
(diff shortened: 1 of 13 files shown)


neilenns (approved):

DocMoebiuz (approved):
@elral Great find
@neilenns Thanks for the review

PR #287, opened 2023-11-27
Double set pin

### LINKED ISSUE #283: Two Pins get set at the same time on Pico

elral:
## Description of changes

Further evaluation of issue #283 showed that PR #285 didn't solve the issue.
Pico is limited to one bank of 16 PWM timer starting at Pin 0. If using a pin > 15 also as PWM output, these pins get doubled.
E.g. using pin 2 as PWM output and unsing pin 18 as PWM output each change on pin 2 or pin 18 will also set the other one.
If one of these pins are used as a digital output everything is fine.

Within the code for the outputs `analogWrite()` is always used toset an output. Doing it this way each pin is used as PWM output for the Pico and the above problem occurs. The only way to solve it is to use only the first 16 pins as PWM output. This change will be done in a separate PR to adapt the `raspberrypi_pico.board.json` by setting `"isPWM": false` for pins > 15. So for pins > 15 the value for setting the pin will either be `0` or `255`.
Within the FW it has to check if the value for each pin is `0` or `255`. In this case `digitalWrite()` is used instead of `analogWrite()` which avoids setting the output as PWM.
This change is only done for the Pico, for AVR's it is kept as it is to not raise another potential problem (which I can not see, just to be absolutely sure).

Fixes #283

### DIFF
--- a/src/MF_Output/MFOutput.cpp
+++ b/src/MF_Output/MFOutput.cpp
@@ -16,26 +16,21 @@ void MFOutput::attach(uint8_t pin)
     _pin   = pin;
 #if defined(ARDUINO_ARCH_RP2040)
     pinMode(_pin, OUTPUT_12MA);
-    digitalWrite(_pin, LOW);
 #else
     pinMode(_pin, OUTPUT);
-    analogWrite(pin, LOW);
 #endif
+    set(LOW);
 }
 
 void MFOutput::set(uint8_t value)
 {
     _value = value;
-#if defined(ARDUINO_ARCH_RP2040)
     if (_value == 0xFF)
         digitalWrite(_pin, HIGH);
     else if (_value == 0x00)
         digitalWrite(_pin, LOW);
     else
         analogWrite(_pin, _value);
-#else
-    analogWrite(_pin, _value);
-#endif
 }
 
 void MFOutput::powerSavingMode(bool state)
--- a/src/MF_Output/MFOutput.cpp
+++ b/src/MF_Output/MFOutput.cpp
@@ -16,21 +16,26 @@ void MFOutput::attach(uint8_t pin)
     _pin   = pin;
 #if defined(ARDUINO_ARCH_RP2040)
     pinMode(_pin, OUTPUT_12MA);
+    digitalWrite(_pin, LOW)
 #else
     pinMode(_pin, OUTPUT);
+    analogWrite(pin, LOW);
 #endif
-    digitalWrite(pin,0);
 }
 
 void MFOutput::set(uint8_t value)
 {
     _value = value;
+#if defined(ARDUINO_ARCH_RP2040)
     if (_value == 0xFF)
         digitalWrite(_pin, HIGH);
     else if (_value == 0x00)
         digitalWrite(_pin, LOW);
     else
         analogWrite(_pin, _value);
+#else
+    analogWrite(_pin, _value);
+#endif
 }
 
 void MFOutput::powerSavingMode(bool state)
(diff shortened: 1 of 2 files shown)


src/MF_Output/MFOutput.cpp
```diff
@@ -14,16 +14,28 @@ MFOutput::MFOutput()
...
     pinMode(_pin, OUTPUT);
-    set(_value);
+    analogWrite(pin, LOW);
 #endif
 }
 
 void MFOutput::set(uint8_t value)
 {
     _value = value;
+#if defined(ARDUINO_ARCH_RP2040)
```
DocMoebiuz, line 29: i think it would be easier to read if we introduce a deviated method like

`setRp2040` and then put the custom logic there instead of having everything inline

src/MF_Output/Output.cpp
```diff
@@ -52,11 +48,21 @@ namespace Output
         // Read led state argument, interpret string as boolean
         int pin   = cmdMessenger.readInt16Arg();
         int state = cmdMessenger.readInt16Arg();
+    
         // Set led
-        analogWrite(pin, state); // why does the UI sends the pin number and not the x.th output number like other devices?
-                                 //  output[pin].set(state);      // once this is changed uncomment this
+#if defined(ARDUINO_ARCH_RP2040)
```
DocMoebiuz, line 53: same here

DocMoebiuz (changes requested):
see my comments - have you thought about it?

src/MF_Output/MFOutput.cpp
elral: Hmhm, in this case the `CommandMessenger.cpp` must also be changed:
```
#if defined(ARDUINO_ARCH_RP2040)
    cmdMessenger.attach(kSetPin, Output::OnSetRP2040);
#else
    cmdMessenger.attach(kSetPin, Output::OnSet);
#endif
```

Due to the differnet pinMode initialization `MFOutput.cpp` would look like:
```
void MFOutput::attach(uint8_t pin)
{
    _pin   = pin;
    pinMode(_pin, OUTPUT);
    analogWrite(pin, LOW);
}

#if defined(ARDUINO_ARCH_RP2040)
void MFOutput::attachRP2040(uint8_t pin)
{
    _pin   = pin;
    pinMode(_pin, OUTPUT_12MA);
    digitalWrite(_pin, LOW);
}
#endif
```

And `Output.cpp` must be as:
```
#if defined(ARDUINO_ARCH_RP2040)
        outputs[outputsRegistered].attachRP2040(pin);
#else
        outputs[outputsRegistered].attach(pin);
#endif
```

Not sure if this increases readability, also as the `CommandMessegner.cpp` must also be changed (changes not as much bundled as before).

DocMoebiuz: i was thinking of a private local method only - nothing fancy

elral: After some more testing the logic from Pico to set outputs can also be used for AVR's.
So I wouldn't differ anymore between both. Just for setting `pinMode()` it has to be differ betwenn AVR and Pico.

The class `MFOutput()` is "only" needed for setting the pinMode and setting the powersave Mode.
I am wondering if we still need this class, both could also be done within `Output.cpp` like setting the output.

DocMoebiuz (approved):

PR #295, opened 2023-12-19
Set single segment on 7 Segment displays

### LINKED ISSUE #294: Set a single Segment on 7 Segment Displays missing
The 7 Segment displays have a fixed character set in the Firmware. Setting an individual segment to show special characters is not possible.

Furthermore the MAX7219 could be used for controlling 64 single LED's connected in a Matrix.

elral:
## Description of changes

An additional function to control a single segment is added. This requires an additional buffer for all Digits for the MAX7219. The already available `rawdata` buffer is still used for the TM1637 display.

The `setSingleSegment(uint8_t subModule, uint8_t segment, uint8_t value)` function requires the subModule, the segment number and the value. Segment numbers are in the range from 0 to 63, 8 bits per Digit, starting with the most right digit. For switching the segment `off` value must be 0x00, all other values are switching the segment `on`.

This can also be used for controlling 64 LED's arranged in a matrix.

Additionally the existing `rawdata` buffer is initialized dynamically and not anymore statically. So the max number of submodules is not fixed anymore to 8 and only the required memory for the number of displays is reserved in the device buffer.

As this is for now only tested with a MAX7219 display, this PR is a draft one until it is tested with TM1637 displays.

Fixes #294


elral:
It's now also tested for the TM1637 displays.

Bitorder is dABCDEFG and from the most right digit to most left digit for both display types (MAX and TM).

E.g. to lit segment `C` from digit `3` set bit (3 \* 8) + 16 = `40` to `1` -> `33,0,0,40,1` would be the command from the connector with module = 0 and submodule = 0.

src/MF_Segment/LedControl_dual.cpp
```diff
@@ -132,15 +133,21 @@ uint8_t LedControl::rawdata[16] = {0};
...
     _clkPin  = clkPin;
     _csPin   = csPin;
 
+    if (!FitInMemory(sizeof(uint8_t) * numDevices * 2))
+        return false;
+    rawdata = new (allocateMemory(sizeof(uint8_t) * numDevices * 2)) uint8_t;
+
     if (isMAX()) {
-        if ((numDevices - 1) > 7) numDevices = 8;
+        if (!FitInMemory(sizeof(uint8_t) * numDevices * 8))
```
DocMoebiuz, line 151:
```suggestion
        // make sure we have max 8 chips in the daisy chain
        if (numDevices > 8) numDevices = 8;

        if (!FitInMemory(sizeof(uint8_t) * numDevices * 8))
```

DocMoebiuz (changes requested):
See comment and suggestion

DocMoebiuz (approved):

PR #296, opened 2023-12-20
Handling Custom Device versions

elral:
## Description of changes

An additional flag `custom_firmware_version =` has been added to the standard environment `[env]` and is used for all standard boards as it gets not overwritten (no additional flag in the boards environment).
The custom device `.ini` files do have also this additional flag, but for each board a different version can be defined (or even the same, but than it should be changed into a common definition).

The `get_version.py` script checks if this additinal flag has a value, and if so this value is used for the firmware version within the filename. Otherwise the built version is used as for the filename as before.

Within `OnGetInfo()` both informations, the new flag and the existing built version, are send back to the connector.
First the new flag is send back as before, so it's still the firmware version. But it could be another one than for the standard boards. Additional the built version is send back which reflects the _core version_ on which the FW is based. This is still required to check which functionalities the board supports (changing name, custom devices, ...)
An example message for `OnGetInfo()` could be:
```
10,Kav Mega,MobiFlight Mega,SN-3ED-CB1,1.0.5,2.5.0;
```
In this case  `custom_firmware_version = 1.0.5` was defined.

If nothing is defined, the response would be:
```
10,Kav Mega,MobiFlight Mega,SN-3ED-CB1,2.5.0,2.5.0;
```

Changes in the connector would also be required to evaluate the additonal message as `core version` to check for functionalities. For now the connector just ignores the additional message, in case of a custom device with a "low" version some functions are not available.

CAUTION! The `get_CustomDevices.py` script is changed temporarily to download the `main' branch from the custom device repo. The changes in this repo are required to get it compiled. Once these changes are part of the next release from the custom device repo, this has to be implemented in this script.

### DIFF
--- a/get_CustomDevices.py
+++ b/get_CustomDevices.py
@@ -3,7 +3,8 @@ import os
 Import("env")
 
 CUSTOMDEVICES_DIR = env.subst("$PROJECT_DIR/CustomDevices")
-CUSTOMDEVICES_TAG = "0.2.0"
+#CUSTOMDEVICES_TAG = "0.2.0"
+CUSTOMDEVICES_TAG = "main"
 
 if not os.path.exists(CUSTOMDEVICES_DIR):
     print ("Cloning Mobiflight-CustomDevices repo ... ")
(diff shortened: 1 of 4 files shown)


elral:
I will close this PR as the handling for custom devices will be done in a complete different way.

PR #306, opened 2024-01-04
Power saving mode

### LINKED ISSUE #299: Propagate PowerSavingMode change to Custom device

elral:
## Description of changes

Adding a function to enable custom devices to enter or leave power saving mode.
If the PowerSavingMode should be entered, a message is sent to the custom device with MessageID = `-2` and Message = `"1"`
If the PowerSavingMode should be leaved, a message is sent to the custom device with MessageID = `-2` and Message = `"0"`

Changes in all custom devices are required to differ between entering and leaving.
This will be done in a separate PR within the CustomDevice Repo.

Fixes #299


src/MF_CustomDevice/CustomDevice.cpp
```diff
@@ -84,4 +84,20 @@ namespace CustomDevice
...
+        This function is called if the status of the PowerSavingMode changes.
+        'state' is true if PowerSaving is enabled
+        'state' is false if PowerSaving is disabled
+        MessageID '-2' for the custom device  for PowerSavingMode
+    ********************************************************************************** */
+    void PowerSave(bool state)
+    {
+        for (uint8_t i = 0; i < customDeviceRegistered; ++i) {
+            if (state)
+                customDevice[i].set(-2, "1");
```
DocMoebiuz: It would be great to have const defines or variables with readable labels for the built-in message ids `-1` and `-2`

DocMoebiuz (changes requested):
see my review comment

src/MF_CustomDevice/CustomDevice.cpp
elral: Changed to `MESSAGEID_POWERSAVINGMODE`

DocMoebiuz: cool! let's do the same for -1

elral: Hmhm, there is no `-1` (Stop of connector)
This message is coming from the connector and directly passed to the custom device.
I see no need to do it here.

DocMoebiuz: i thought it would make it more obvious that it's an internal event. Ideally stop would also come from the firmware as we discussed recently. Stop as message on Board level. On the other hand it's also good to have some flexibility on the App side  

Since this is a special event I would still give it a speaking name.

elral: Hmhm, I already changed the `-1` to `MESSAGEID_POWERSAVINGMODE`.
Or do you mean something different?

DocMoebiuz (approved):

PR #307, opened 2024-01-05
HIGH-LOW logic configurable via build flag

### LINKED ISSUE #302: Make HIGH-LOW logic configurable via build flag
Currently, MobiFlight treats HIGH as 255 and LOW as 0 for outputs. There are circuits where this should be the other way around.

elral:
## Description of changes

Via the configuration flag `REVERSED_OUTPUT_OUTPUT` within the platformio.ini outputs can be set to a reversed output logic, means setting the ouput from the conenctor will set the output to `LOW` and vice versa.
The same applies to the outputs from the outputshifter via the configuration flag `REVERSED_OUTPUT_OUTPUTSHIFTER`.

Both configuration flags are disabled in the platformio.ini. To use the logic uncomment one of them or both.

Fixes #302


src/MF_OutputShifter/MFOutputShifter.cpp
```diff
@@ -19,7 +19,11 @@ void MFOutputShifter::setPin(uint8_t pin, uint8_t value, uint8_t refresh)
     uint8_t idx = (pin & 0xF8) >> 3;
     uint8_t msk = (0x01 << (pin & 0x07));
 
+#ifdef REVERSED_OUTPUT_OUTPUTSHIFTER
+    if (value == 0) {
+#else
     if (value > 0) {
```
DocMoebiuz: Could this not be like

```cpp
if (value != MF_LOW)
```

And we would not need the `#ifdef` `#else`

DocMoebiuz (changes requested):

src/MF_OutputShifter/MFOutputShifter.cpp
elral: Hmhm, MF_LOW is only defined for the outputs within MFOutput.h.

Additionally I just recognized that the clear() function must also be adapted

elral: Yeah, it makes absolutely sense.
Added the defines to MFOutputShifter.h, otherwise I had to include MFOutput.h or add the defines to all MFBoards.h (which might be also a solution)

elral:
Within `setPin()` this `if (value != MF_LOW)` is used in this PR.
But this does not work for reversed outout logic.
For reversed output logic `MF_LOW ` is set to `0x01`, but it has to be checked for all values > 0.

Logic has to be reworked before merging!

DocMoebiuz (approved):
Awesome!

PR #309, opened 2024-01-10
Add core version

### LINKED ISSUE #298: Provide a solid workflow for creating a custom board with custom devices
First thoughts

image

elral:
## Description of changes

Custom Devices will be build within there own Repo in the short future. The will inform the connector about there firmware version and on which core version (`core_firmware_version`) they are build.

This Repo will build only the core boards in the near future, so both versions are always the same.
Built from custom devices from this Repo will also report the same FW and Core version, like it is the actual status.

Fixes #298


DocMoebiuz (approved):
Thanks!

DocMoebiuz (approved):

PR #311, opened 2024-01-14
Custom Devices will not be built anymore

### LINKED ISSUE #310: Own Repos for Custom Devices
The existing custom devices are transferred to single Community Repos based on a Template which was set up in the last weeks.

No need anymore to build the Custom Boards out of this Repo.
Only core boards should be build.

elral:
## Description of changes

Custom devices will not be build anymore out of this repo. Only FW for core boards will be built.
The download script is deleted, the platformio.ini does not include the build definitions from the custom devices and the Action for github will not download the custom devices anymore.

Fixes #310


DocMoebiuz (approved):
Perfect!

PR #312, opened 2024-01-14
Don't delete config if UniqueID was used for AVR's

### LINKED ISSUE #313: Config gets deleted if UniqueID was used for AVR's
On a short period of time UniqueID was used for AVR's. This was rolled back and it was intended to generate a new serial number. Unfortenutely the config gets also deleted.

elral:
## Description of changes

On a short period of time UniqueID was used for AVR's. This was rolled back and it was intended to generate a new serial number. Unfortenutely this wasn't introduced, instead a new board was detected and the config was deleted by writing 0x00 to the first location of the config in the EEPROM.
With this a serial number gets generated if a UniqueID is used for AVR's. Additionally it's checked if it's a first time starting up this board by checking the first eeprom loation for the config. It must be 0xFF (blank EEPROM), otherwise it is assumed that reading the serial number failed. In this case no new serial number is generated.
Deleting the config will not be done anymore.

On a first start up of the board the EEPROM contains always 0xFF. If the config get's read from the connector, all 0xFF up to the MEMLEN_CONFIG gets transferred. This is changed, it is checked if the first Byte of the config is 0xFF, in this case an empty config is assumed and no config gets transferred to the connector.

Fixes #313

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -88,7 +88,7 @@ bool readConfigLength()
     configLength        = 0;
 
     if (MFeeprom.read_byte(MEM_OFFSET_CONFIG) == 0xFF)
-        return;
+        return false;
     while (MFeeprom.read_byte(addreeprom++) != 0x00) {
         configLength++;
         if (addreeprom > length) {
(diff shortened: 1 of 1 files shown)


DocMoebiuz (approved):
Thanks

PR #320, opened 2024-03-22
Custom Boards - 2nd core can be used on rp2040

### LINKED ISSUE #319: Community devices might need too much time for setting new value
With more and more community devices upcoming and using OLED's there could be a timing problem if the set() function from the custom devices take a longer time. "Bigger" OLED's may need so much time, that e.g. encoder events could be missed.

elral:
## Description of changes

On Picos the 2nd core is used to run the `set()` function of the community devices.
With this change setting a new value to a community device will not lead to missed input events even if the `set()` function needs much more time.

The `FreeRTOS` lib is used to initialize and run the 2nd core of a Pico. Data from the 1st core to the 2nd core is transferred by using the build in `fifo`. Once all data is received the 2nd core calls the `set()` function of the community device.
To activate this feature only one define within `custom.platformio.ini` is required. Once `-DUSE_2ND_CORE` is set, this feature is enabled. If it is not set or even not available, the 2nd core of the pico is not used and everything is like before. So it's fully backward compatible, the 2nd core will not be used as long as the above mentioned flag is not set.

Fixes #319


src/MF_Stepper/Stepper.cpp
```diff
@@ -110,4 +148,45 @@ namespace Stepper
...
+{
+    rp2040.fifo.push(true); // inform core 1 to be ready
+}
+
+void loop1()
+{
+    uint8_t  command, stepper;
+    int32_t param1, param2;
+
+    while (1) {
```
DocMoebiuz, line 169: while (true) would be prettier.

DocMoebiuz:
i think the description of this PR requires a bit more documentation on the actual scope and how to use it.

* Enable 2nd core support for Custom Devices
* Enable 2nd core support for Steppers
* Default settings used for firmware build

DocMoebiuz (changes requested):
see my comments.

DocMoebiuz (approved):

PR #322, opened 2024-05-12
Fixed warning during compile for community devices

elral:
## Description of changes

When compiling the sources for community devices a warning is generated.
This PR fixes the warnings.


DocMoebiuz (approved):

DocMoebiuz (commented):
Should this not be using the F() macro?

elral:
> Should this not be using the F() macro?

Oh yes.

/edit: NO, does not work with the macro F()

PR #327, opened 2024-07-06
Read config from EEPROM only if no config is in Flash

### LINKED ISSUE #326: Config could get read doubled
With actual main branch (pre 2.5.2) a configuration can be defined within the flash memory.
If this is the case, no new configuration from the connector get's saved to the EEPROM.
This is as intended, a mixed configuration from EEPROM and flash is not foreseen.

But in case of an updated of an already existing FW and configuration in the EEPROM with a version with a configutation within the flash, both versions gets reported to the connector and gets loaded.
This should not be the case, only the version from flash should get loaded and reported to the connector.

elral:
## Description of changes

If a configuration is already saved in the EEPROM, and the board gets updated with a new Firmware where the configuration is part of the flash, the exsiting configuration in the EEPROM gets ignored and only the one from the flash gets reported to the connector and loaded into the internal memory.

Before sending the configuration to the connector, it gets's checked if a configuration is stored in the flash. If this is the case (configLength > 0), this configuration is send to the connector. If no config in the flash is available (configLength = 0) but a config in the EEPROM saved, this config gets send to the connector.
The same applies for loading the config into the internal memory.

For now it is only checked if a configuration in the EEPROM `OR` Flash is available. If a configuration is saved, this one gets reported to the connector and loaded into the internal memory.

Fixes #326

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -130,7 +130,8 @@ void OnSetConfig()
     char   *cfg    = cmdMessenger.readStringArg();
     uint8_t cfgLen = strlen(cfg);
 
-    if (configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG) {
+   bool maxConfigLengthNotExceeded = configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG;
+    if (maxConfigLengthNotExceeded) {
         MFeeprom.write_block(MEM_OFFSET_CONFIG + configLengthEEPROM, cfg, cfgLen + 1); // save the received config string including the terminatung NULL (+1) to EEPROM
         configLengthEEPROM += cfgLen;
         cmdMessenger.sendCmd(kStatus, configLengthEEPROM);
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -84,6 +84,14 @@ void resetConfig();
 void readConfig();
 void _activateConfig();
 void readConfigFromMemory(bool configFromFlash);
+bool configStoredInFlash()
+{
+    return configLengthFlash > 0;
+}
+bool configStoredInEEPROM()
+{
+    return configLengthEEPROM > 0;
+}
 
 // ************************************************************
 // configBuffer handling
@@ -130,7 +138,7 @@ void OnSetConfig()
     char   *cfg    = cmdMessenger.readStringArg();
     uint8_t cfgLen = strlen(cfg);
 
-   bool maxConfigLengthNotExceeded = configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG;
+    bool maxConfigLengthNotExceeded = configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG;
     if (maxConfigLengthNotExceeded) {
         MFeeprom.write_block(MEM_OFFSET_CONFIG + configLengthEEPROM, cfg, cfgLen + 1); // save the received config string including the terminatung NULL (+1) to EEPROM
         configLengthEEPROM += cfgLen;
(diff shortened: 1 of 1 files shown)


elral:
For backward compatibility it is first checked, if a config in EEPROM is available. If so, this config will still be used.
Otherwise it is checked, if a config in Flash is available and if so this config get's used.

It is still possible, to upload a user config even if a config in flash is available. This overrules the flash config.
But this enables also the user, to use the flash config if an empty config get's uploaded.
The only drawback is that the connector has to be restarted if an empty config is uploaded and the flash config get's be used because the config is not read again after an upload.

src/Config.cpp
```diff
@@ -122,22 +122,20 @@ void OnSetConfig()
...
-        uint8_t cfgLen = strlen(cfg);
-
-        if (configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG) {
-            MFeeprom.write_block(MEM_OFFSET_CONFIG + configLengthEEPROM, cfg, cfgLen + 1); // save the received config string including the terminatung NULL (+1) to EEPROM
-            configLengthEEPROM += cfgLen;
-            cmdMessenger.sendCmd(kStatus, configLengthEEPROM);
-        } else
-            cmdMessenger.sendCmd(kStatus, -1); // last successfull saving block is already NULL terminated, nothing more todo
-    } else {
+    // A config can be in flash or in EEPROM, bur only one must be used
```
DocMoebiuz:
```suggestion
    // A config can be in flash or in EEPROM, but only one option must be used
```

src/Config.cpp
```diff
@@ -122,22 +122,20 @@ void OnSetConfig()
...
+    // Once a config is in EEPROM, this config will be loaded and reported to the connector
+    // If no config is in EEPROM, the config from flash will be used if available
+    // This ensures backwards compatibility if a board gets updated with a config in flash
+    // but also have a user config in EEPROM
+    char   *cfg    = cmdMessenger.readStringArg();
+    uint8_t cfgLen = strlen(cfg);
+
+    if (configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG) {
```
DocMoebiuz:
```suggestion
   bool maxConfigLengthNotExceeded = configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG;
    if (maxConfigLengthNotExceeded) {
```

src/Config.cpp
```diff
@@ -352,17 +350,15 @@ void InitArrays(uint8_t *numberDevices)
...
     }
     InitArrays(numberDevices);
     if (configLengthFlash > 0) {
         readConfigFromMemory(true);
-    }
-    if (configLengthEEPROM > 0) {
+    } else if (configLengthEEPROM > 0) {
         readConfigFromMemory(false);
     }
 }
```
DocMoebiuz: I would like to avoid so many if-else blocks
Suggestion:

``` cpp
void readConfig()
{
    uint8_t numberDevices[kTypeMax] = {0};

    bool useFlash = configLengthFlash > 0;
    bool useEEPROM = configLengthEEPROM > 0;

    // Early return if no valid configuration is found
    if (!useFlash && !useEEPROM) {
        InitArrays(numberDevices);
        return;
    }

    // Determine which configuration to use and proceed
    GetArraySizes(numberDevices, useFlash);
    InitArrays(numberDevices);
    readConfigFromMemory(useFlash);
}
```

src/Config.cpp
```diff
@@ -564,23 +560,17 @@ void readConfigFromMemory(bool configFromFlash)
...
     if (configLengthEEPROM > 0) {
-        if (sentFromFlash)
-            cmdMessenger.sendArg((char)MFeeprom.read_byte(MEM_OFFSET_CONFIG));
-        else
-            cmdMessenger.sendCmdArg((char)MFeeprom.read_byte(MEM_OFFSET_CONFIG));
+        cmdMessenger.sendCmdArg((char)MFeeprom.read_byte(MEM_OFFSET_CONFIG));
         for (uint16_t i = 1; i < configLengthEEPROM; i++) {
             cmdMessenger.sendArg((char)MFeeprom.read_byte(MEM_OFFSET_CONFIG + i));
         }
+    } else if (configLengthFlash > 0) {
```
DocMoebiuz: since now i am seeing this over and over again, what about extracting this into a method: 

``` cpp
bool configStoredInFlash() {
  configLengthFlash > 0
}
```

Would improve readability in my opinion.

DocMoebiuz (changes requested):

DocMoebiuz (approved):

src/Config.cpp
elral: Ups, `InitArrays(numberDevices);` is not required for the early return. But as the number of all devices is set to `0`, nothing worse happens.

PR #329, opened 2024-08-07
Stop 2nd core for pico during config upload

### LINKED ISSUE #328: Config upload  sometimes doesn't work for Pico
Uploading a config to the pico doesn't work sometimes.
The connector reports an error uploading the config because the pico crashes in this case.
This does not happen always, somtimes it's working.

elral:
## Description of changes

If a custom device is defined which is running on the 2nd core from the Pico and this custom device has regular updates, the Pico crashes sometimes while uploading a new config file.
To fix this issue the update() function on the 2nd core has to be stopped before clearing the custom devices and uploading a new config file. After clearing the custom devices the flag is resetted. As no custom devices are defined anymore the update() function gets not called anymore and a new config file can be uploaded.

When uploading a new config, this config gets saved in the EEPROM. If the 2nd core of the Pico is used, this one has to be paused to write the changes to the Flash (EEPROM emulation). Afterwards the 2nd core can be run again.

A new function within CustomDevice.cpp is added. This function sends a message to the 2nd core of the Pico to stop updating the custom device. The function waits for a response from the 2nd core via the FiFo buffer.

Within the loop1() function it is checked, if a `stop updating` command is received, If so a flag is set and the custom device gets not updated anymore. Also it is checked if a `start updating` command is received and if so, the custom device can be updated again. In both cases the 1st core gets informed.

Within Config.cpp the `stop updating` function is called before clearing the custom devices. Afterwards the `start updating` function is called to enable updating custom devices again.

Fixes #328

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -173,7 +173,13 @@ void resetConfig()
     DigInMux::Clear();
 #endif
 #if MF_CUSTOMDEVICE_SUPPORT == 1
+#if defined(USE_2ND_CORE)
+    CustomDevice::stopUpdate2ndCore(true);
+#endif
     CustomDevice::Clear();
+#if defined(USE_2ND_CORE)
+    CustomDevice::stopUpdate2ndCore(false);
+#endif
 #endif
     configLengthEEPROM = 0;
     configActivated    = false;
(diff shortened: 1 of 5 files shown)


DocMoebiuz (approved):

DocMoebiuz:
@elral is there any docs available about this feature? The description explains internal workings and that's great - but what would be relevant for the (custom device) developer?

PR #333, opened 2024-09-05
write eeprom or name if they are available even config is in flash

elral:
## Description of changes

In addition to PR #327 fully backwards compatibility is now ensured.
Once a config is in EEPROM, this config can still be changed even a config is in flash.
Once a name is in EEPROM, this name can still be changed even a config is in flash.
If no name and no config is in EEPROM, name and config can not be changed anymore.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -79,7 +79,6 @@ uint16_t       configLengthEEPROM              = 0;
 boolean        configActivated                 = false;
 uint16_t       pNameBuffer                     = 0; // pointer for nameBuffer during reading of config
 const uint16_t configLengthFlash               = sizeof(CustomDeviceConfig);
-// if config is in EEPROM, ensure upload a new config even if a config in flash is available
 
 void resetConfig();
 void readConfig();
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -79,11 +79,14 @@ uint16_t       configLengthEEPROM              = 0;
 boolean        configActivated                 = false;
 uint16_t       pNameBuffer                     = 0; // pointer for nameBuffer during reading of config
 const uint16_t configLengthFlash               = sizeof(CustomDeviceConfig);
+// if config is in EEPROM, ensure upload a new config even if a config in flash is available
+bool configEEPROMavailable = false;
 
 void resetConfig();
 void readConfig();
 void _activateConfig();
 void readConfigFromMemory(bool configFromFlash);
+
 bool configStoredInFlash()
 {
     return configLengthFlash > 0;
@@ -112,6 +115,7 @@ bool readconfigLengthEEPROM()
             return false;
         }
     }
+    configEEPROMavailable = true;
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


src/Config.cpp
```diff
@@ -718,9 +728,10 @@ void OnGenNewSerial()
 // ************************************************************
 void storeName()
 {
-    if (!configStoredInFlash()) {
+    if (configStoredInEEPROM() || !configStoredInFlash()) {
         MFeeprom.write_byte(MEM_OFFSET_NAME, '#');
         MFeeprom.write_block(MEM_OFFSET_NAME + 1, name, MEM_LEN_NAME - 1);
+        // MFeeprom.commit() is not required, name is always set before config
```
DocMoebiuz, line 734: better remove unnecessary lines

src/Config.cpp
```diff
@@ -718,9 +728,10 @@ void OnGenNewSerial()
 // ************************************************************
 void storeName()
 {
-    if (!configStoredInFlash()) {
+    if (configStoredInEEPROM() || !configStoredInFlash()) {
```
DocMoebiuz: is this case not the case when !configStoredInFlash()?

src/Config.cpp
```diff
@@ -735,7 +746,7 @@ void restoreName()
 void OnSetName()
 {
     char *cfg = cmdMessenger.readStringArg();
-    if (!configStoredInFlash()) {
+    if (configStoredInEEPROM() || !configStoredInFlash()) {
```
DocMoebiuz: is this case not the case when !configStoredInFlash()?

src/Config.cpp
```diff
@@ -138,16 +140,24 @@ void OnSetConfig()
...
-        cmdMessenger.sendCmd(kStatus, -1); // last successfull saving block is already NULL terminated, nothing more todo
+    if (configStoredInEEPROM() || !configStoredInFlash()) {
+        bool maxConfigLengthNotExceeded = configLengthEEPROM + cfgLen + 1 < MEM_LEN_CONFIG;
+        if (maxConfigLengthNotExceeded) {
+            // save the received config string including the terminatung NULL (+1) to EEPROM
+            MFeeprom.write_block(MEM_OFFSET_CONFIG + configLengthEEPROM, cfg, cfgLen + 1);
+            configLengthEEPROM += cfgLen;
+            cmdMessenger.sendCmd(kStatus, configLengthEEPROM);
+        } else
+            // staus message to connector, failure on setting config
```
DocMoebiuz, line 150: With these multilines in the else block better use `{` `}`

DocMoebiuz (changes requested):
only small notes and a question which use case that is where it can be 

`if (configStoredInEEPROM() || !configStoredInFlash())`

src/Config.cpp
elral: I see this more as an important comment for two reasons:
* users which are not familiary with the code may ask why the name gets not commited to the EEPROM
* should the connector differ between saving the config the name (I don't see a reason to save always the name even it is unchanged) in the future, commiting has to be done. This comment is to remember (hopefully) this.

src/Config.cpp
```diff
@@ -138,16 +140,24 @@ void OnSetConfig()
...
+            // connector does not check for status = -1
+            cmdMessenger.sendCmd(kStatus, -1);
```
elral, line 152: For now the connector reports a successfull upload of the config, even if `-1` is send back.
If no message is send back, the connector shows an error message (due to time out) that the config uploads fails.

Does it make sense to uncomment `cmdMessenger.sendCmd(kStatus, -1);` as long as the connector does not handle this?
There were discussion before to implement a `kError` message which would be sensefull also in this case ;)

DocMoebiuz (approved):

PR #336, opened 2024-09-26
support for multiple configuration lines on same display

niddu85:
## Description of changes

with this small change is possible to define multiple configuration lines on Mobiflight Connector in order to manage different parts of the display on different configuration lines. 
It's enough to put a "^" character in the display to skip that character from being shown on the display

Line 1
image

Line 2
image

Result
PXL_20240926_205308026

### DIFF
--- a/src/MF_LCDDisplay/MFLCDDisplay.cpp
+++ b/src/MF_LCDDisplay/MFLCDDisplay.cpp
@@ -15,9 +15,17 @@ void MFLCDDisplay::display(const char *string)
 {
     if (!_initialized)
         return;
-    for (uint8_t line = 0; line != _lines; line++) {
-        _lcdDisplay.setCursor(0, line);
-        _lcdDisplay.writeString(&string[line * _cols], _cols);
+    
+    for (uint8_t line = 0; line != _lines; line++) 
+    {
+        for(uint8_t col = 0; col != _cols; col++)
+        {
+            _lcdDisplay.setCursor(col, line);
+            if(string[(line * _cols)+col] != '^')
+            {
+                _lcdDisplay.write(string[(line * _cols)+col]);
+            }
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


DocMoebiuz:
It's an interesting idea... but we are using the caret in actual configs, for example to indicate the transponder digit which will be modified when using a encoder.
Don't forget that we have placeholders that are user defined and could conflict with this function.

So how would it work for these configs? 

I would also expect that a label explains such behavior, compare to the text that we have at the moment.

niddu85:
I agree with your comment and I updated the PR adding a configurable escape character on the Connector, that is translated into a non-printable character (character code 28) and used by the display to skip the character being overwritten.

This PR should be used together with:
 
https://github.com/MobiFlight/MobiFlight-Connector/pull/1829

niddu85:
Hi maintainers,
Is this feature interesting to be merged or shall I close the PR?
I'm using it since I created it with no issues.

B.R.

elral:
For now, everything which gets displayed, is in one config (which I personally prefer). With your proposal that would be splitted up into two or more configs if I understand it correct.

I still do not see the benefit for this as at least everything can also be done in the existing way and it makes more diffucult to keep the overview.

@DocMoebiuz what do you think about this?

niddu85:
The benefit I got is the following: 
I made a panel with 2 rotator switches with 7 positions each.
PXL_20241207_075954087 MP

Each switch controls what's displayed on it's line ( the top switch controls the first 2 rows of the display and the bottom switch the last 2 rows)
If I had to create the configuration in the actual way I had to create 49 (7x7) configuration lines.
With my feature only 14 lines are needed.

Let me attach you my configuration to let you better understand 
[MobiFlight config.zip](https://github.com/user-attachments/files/18047677/MobiFlight.config.zip)

PR #338, opened 2024-10-18
Fixes analog read for pins A6 and A7

### LINKED ISSUE #337: Issues A6 and A7 with potentiometers on ProMicro
Hello,

We are having issue using pins A6 and A7 on promicro board.

That is the current situation trying every analog pin with a potentiometer :

A0 works
A1 works
A2 works
A3 works, but *also* shows on 4 (A6)
4 shows as 6 (A7)
6 does not work 
8 (A8) works 
9 (A9) works

This subject has been discussed on discord [here](https://discord.com/channels/608690978081210392/1293980096926846976) and is discussed [here](https://discord.com/channels/608690978081210392/1296218159463792681/1296218161217146931).

Many thanks !

[523a1765757b7f5c6e8b4567](https://github.com/user-attachments/assets/be8f4445-d1a7-4bf3-8443-9e2c2db01ee7)

elral:
## Description of changes

ProMicro has some special pin mapping within pins_arduino.h through which analog pins `A6` and `A7` can not be adressed via there digital pin numbers. But Mobiflight does only use the digital pin numbers also for pins with analog readin function.
With this change it is checked if the digitla pin number is one of the two pin numbers and in this case a reampping is done so that the pin number fits to the required one.

Fixes #337

### DIFF
--- a/src/MF_Analog/MFAnalog.cpp
+++ b/src/MF_Analog/MFAnalog.cpp
@@ -18,6 +18,12 @@ void MFAnalog::attach(uint8_t pin, const char *name, uint8_t sensitivity)
     _sensitivity = sensitivity;
     _pin         = pin;
     _name        = name;
+#if defined(ARDUINO_AVR_PROMICRO16)
+    if (_pin == 4)
+        _pin = A6;
+    else if (_pin == 6)
+        _pin = A7;
+#endif
     pinMode(_pin, INPUT_PULLUP); // set pin to input. Could use OUTPUT for analog, but shows the intention :-)
     // Fill averaging buffers with initial reading
     for (uint8_t i = 0; i < ADC_MAX_AVERAGE; i++) {
--- a/src/MF_Analog/MFAnalog.cpp
+++ b/src/MF_Analog/MFAnalog.cpp
@@ -19,6 +19,9 @@ void MFAnalog::attach(uint8_t pin, const char *name, uint8_t sensitivity)
     _pin         = pin;
     _name        = name;
 #if defined(ARDUINO_AVR_PROMICRO16)
+    // ProMicro has a special pin assignment for analog pins
+    // therefore reading from A6 and A7 does not work
+    // via "digital" pins. See also pins_arduino.h
     if (_pin == 4)
         _pin = A6;
     else if (_pin == 6)


src/MF_Analog/MFAnalog.cpp
```diff
@@ -18,6 +18,15 @@ void MFAnalog::attach(uint8_t pin, const char *name, uint8_t sensitivity)
     _sensitivity = sensitivity;
     _pin         = pin;
     _name        = name;
+#if defined(ARDUINO_AVR_PROMICRO16)
+    // ProMicro has a special pin assignment for analog pins
+    // therefore reading from A6 and A7 does not work
+    // via "digital" pins. See also pins_arduino.h
+    if (_pin == 4)
+        _pin = A6;
```
DocMoebiuz, line 26: Alternatively we could have mapped the label A6 in the boards.json file, no? 
That would leave this special treatment out here. My first thought was, that special handling for certain variants is not so great. i know we have done it for Pico, but that's an entirely different platform.

DocMoebiuz (commented):
Can we chat about this change and see if we want to keep it here?

elral:
Summary from our chat on Discord:

_Analog A6 is on pin 4. To use this pin 4 as analog in, the Arduino framework expects analogRead(A6). A6 is defined as 24 in the .h file. The analogRead() function subtracts 18 for pin numbers > 18. So I can also specify the value 6 as the pin number. But that is not pin 4 that I want to connect the potentiometer to._

_Ax (x-18 if x>17) is equivalent to digital pin x, except for A6 and A7. Therefore you have to check if it is one of these two pins and then map it._

_This special feature, that the digital pin number cannot be used for analogRead(), only exists with these two pins on the ProMicro._

_I see no chance just to do it in the board.json file._

See also from arduino\_pins.h:

```
// Mapping of analog pins as digital I/O
// A6-A11 share with digital pins
static const uint8_t A0 = 18;
static const uint8_t A1 = 19;
static const uint8_t A2 = 20;
static const uint8_t A3 = 21;
static const uint8_t A4 = 22;
static const uint8_t A5 = 23;
static const uint8_t A6 = 24;   // D4
static const uint8_t A7 = 25;   // D6
static const uint8_t A8 = 26;   // D8
static const uint8_t A9 = 27;   // D9
static const uint8_t A10 = 28;  // D10
static const uint8_t A11 = 29;  // D12

#define analogPinToChannel(P)  ( pgm_read_byte( analog_pin_to_channel_PGM + (P) ) )

//  _AVR_ATmega32U4 has an unusual mapping of pins to channels
extern const uint8_t PROGMEM analog_pin_to_channel_PGM[];
const uint8_t PROGMEM analog_pin_to_channel_PGM[] = {
    7,  // A0               PF7                 ADC7
    6,  // A1               PF6                 ADC6
    5,  // A2               PF5                 ADC5
    4,  // A3               PF4                 ADC4
    1,  // A4               PF1                 ADC1
    0,  // A5               PF0                 ADC0
    8,  // A6       D4      PD4                 ADC8
    10, // A7       D6      PD7                 ADC10
    11, // A8       D8      PB4                 ADC11
    12, // A9       D9      PB5                 ADC12
    13, // A10      D10     PB6                 ADC13
    9   // A11      D12     PD6                 ADC9
};
```

Are there anymore informations requuired to merge this Pullrequest?

DocMoebiuz (approved):

PR #339, opened 2024-10-21
Add power saving mode for outputsifter

### LINKED ISSUE #301: Add PowerSavingMode to Output Shift Register

elral:
## Description of changes

Power Saving mode is added for outputshifters.

Once the power saving mode should be entered, all pins of the outputshifters are set to `0` or `1` on case of reversed polarity.
When the power saving mode should be leaved, the actual status is shifted out and the pins are set to the original status.

Fixes #301


DocMoebiuz (approved):
Great!

PR #342, opened 2024-11-01
Set FW version to 3.0

elral:
## Description of changes

Next FW version will be 3.0 instead of 2.5.2. This is changed in all board.json files.
It overrules #335 (already merged) where it was intended to have version 2.5.2.


DocMoebiuz (changes requested):
Change to 3.0.0 please

DocMoebiuz (approved):

PR #343, opened 2024-11-01
Cleanup all source files

elral:
## Description of changes

This is a major PR but it has only editorial changes. Functionality is **not** changed, tested by compiling this branch and the main branch. For all boards the hex size hasn't changed.

1. All files are opend and autoformatted. So each file is affected by this PR
2. Most files have included `mobiflight.h` which included severall more `.h` files. For most files, all include files are not required. So `mobiflight.h` is deleted and in each other file only the required include files are included.
3. Board definition is done in two different locations, in `MF_Boards.h` for each board and `platformio.ini`. With this PR all board definitions are moved to `platformio.ini` and all `MF_Boards.h` files are deleted. All `board.json` files and `reset.hex` files are now in one folder. Therefore the `copy_fw_files.py` is also adapted.
4. Mixed usage of `ifdef` and `if defined()` is commonly changed to `if defined()`.


DocMoebiuz (changes requested):
Change all 

#if defined(MF_OUTPUT_SHIFTER_SUPPORT)

to 

#ifdef MF_OUTPUT_SHIFTER_SUPPORT

Find: ```#if\s*defined\s*\(\s*([A-Za-z0-9_]+)\s*\)```
Replace ```#ifdef $1``` or ```#ifdef \1```

DocMoebiuz (approved):

PR #348, opened 2025-01-06
init analog input w/o pullup

### LINKED ISSUE #347: Analog read shows non-linear behaviour
Using a potentiometer for reading the analog value shows that the values are not linear compared to the position of the potentiometer.
This was reported on [discord](https://discord.com/channels/608690978081210392/1319802369839464550)

In the actual code a pin for reading analog values is initialized with pullups enabled. When connecting a potentiometer to the pin this internal resistor is always parallel to the "upper" part of the potentiometer. The internal pullup resistor is ~35kOhm.
The user who reported this issue used a 500kOhm potentiometer. If the pot is on the "upper" or "lower" end, the correct value is measured. An example for the mid position would be:
"upper" and "lower" part of the potentiometer is 250kOhm, parallel to the "upper" part is 35kOhm, the resulting value is 31kOhm which far away from the 250kOhm for mid position and a high non linearity.
For a 10kOhm Potentiometer it's quite better, the deviation in the mid position is "only" 3%. But it could be still visible espacially for closed loop controls of motorizes axis.

elral:
## Description of changes

In the actual code a pin for reading analog values is initialized with pullups enabled. When connecting a potentiometer to the pin this internal resistor is always parallel to the "upper" part of the potentiometer. The internal pullup resistor is ~35kOhm.
This results in a non linear behaviour of the potentiometer. For a 10kOhm potentiometer the deviation is max. 4% at 70% position, this increases for higher values of the potentiometer.
Examples are:
50kOhm: 16% deviation at 70% position
100kOhm: 26% deviation at 70% position
500kOhm: 56% deviation at 80% position

The code change is not to use the internal pullup resistor. Also setting the `pinMode()` is not required as this is done during the `analogRead()` function.

Fixes #347

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -514,10 +514,14 @@ void readConfigFromMemory(bool configFromFlash)
 #endif
 
 #if MF_ANALOG_SUPPORT == 1
+        case kTypeAnalogInputDeprecated:
         case kTypeAnalogInput:
             params[0] = readUint(&addrMem, configFromFlash);                              // pin number
             params[1] = readUint(&addrMem, configFromFlash);                              // sensitivity
-            Analog::Add(params[0], &nameBuffer[pNameBuffer], params[1]);                  // MUST be before readName because readName returns the pointer for the NEXT Name
+            if (command == kTypeAnalogInputDeprecated)
+                Analog::Add(params[0], &nameBuffer[pNameBuffer], params[1], true);        // MUST be before readName because readName returns the pointer for the NEXT Name
+            else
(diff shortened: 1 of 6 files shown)


DocMoebiuz (changes requested):
We want to introduce a new device type so that 
* old, existing analog inputs maintain the current logic (with pullup)
* newly added analog inputs will use no pullup

elral:
With last commit a new device type is added to keep backwards compatibility

DocMoebiuz (approved):

PR #362, opened 2025-08-13
Fix powersaving mode

### LINKED ISSUE #361: Timer for Power Safe Mode does not get resetted
Messages to output devices do not reset the timer for entering the Power Safe Mode.
If a message to an output device is send, this message should also reset the timer for entering the power safe mode.
Espacially in Test Mode keep alive messages are not send, so output devices are entering the power safe mode even output messages are generated.

elral:
## Description of changes

If a message to an output device is send, this message should also reset the timer for entering the power safe mode.
Espacially in Test Mode keep alive messages are not send, so output devices are entering the power safe mode even output messages are generated.

Messages to output devices will now reset the timer for entering the Power Safe Mode.
A wrapper function for each output device is added which is called when a message is received.
This function resets the timer for entering the output device and calls the `OnSet()` function accordingly.

Fixes #361

### DIFF
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -92,6 +92,14 @@ void PowerSavingWrapper_OutputShifter_OnSet()
 }
 #endif
 
+#ifdef MF_SERVO_SUPPORT
+void PowerSavingWrapper_Servo_OnSet()
+{
+    trackCommand();
+    Servos::OnSet();
+}
+#endif
+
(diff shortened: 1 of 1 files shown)


elral:
The connector will send a keep alive message also in `Test` mode.
No changes in the FW are required.

See [Issue](https://github.com/MobiFlight/MobiFlight-Connector/issues/2297)

PR #364, opened 2025-08-20
Send command 'kRetriggerDone' after retrigger of all inputs

### LINKED ISSUE #363: Missing messages during retrigger
After connect the connector requests a retrigger of all inputs if this option is choosed.
With big configuratons the messages the FW sends could be quite much. If it takes too much time, the connector sends the first messages to output devices even the retrigger has not finished. This could lead to lost messages and some outputs will not get set.

See also Issue (_will be added_) from the connector repo.

elral:
## Description of changes

A command gets send when all retrigger messages has been send.
This avoids lost messages from the connector if quite a lot events from the FW are send due to a big configuration.

Fixes #363

### DIFF
--- a/src/CommandMessenger.cpp
+++ b/src/CommandMessenger.cpp
@@ -147,6 +147,7 @@ void OnTrigger()
 #ifdef MF_ANALOG_SUPPORT
     Analog::OnTrigger();
 #endif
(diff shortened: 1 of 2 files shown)


DocMoebiuz (approved):
Nice!

PR #367, opened 2025-09-20
Add ESP32 Support

miathedev:
**NOTE**
> Please dont merge it yet cause theres an ongoing MR pending to make this work: https://github.com/MobiFlight/LiquidCrystal_I2C/pull/4

> This is an ongoing MR

This pull request adds initial support for the ESP32-S3 board (specifically the Heltec Wireless Stick V3) to the project. The changes include board configuration, build system updates, and multiple code modifications to ensure compatibility with the ESP32/ESP32-S3 architecture. The most significant updates are grouped below.

**ESP32-S3 Board Support**

* Added a new board definition file for the Heltec Wireless Stick V3 (`_Boards/esp32s3.board.json`), specifying hardware IDs, pin mappings, module limits, and other board-specific settings.
* Updated `platformio.ini` to add a new build environment (`[env:mobiflight_esp32s3]`) for the ESP32-S3, including necessary build flags and dependencies.

**Codebase Compatibility Adjustments**

* Updated memory allocation logic in `allocateMem.cpp` and `allocateMem.h` to use `uint8_t` buffers and pointers for both AVR and ESP32 architectures, ensuring consistent memory handling.
**Peripheral and Pin Handling for ESP32**

* Updated analog pin handling in `MFAnalog.cpp` to clarify ESP32's analog pin usage and allow for future mapping adjustments.
* Updated output pin initialization in `MFOutput.cpp` to ensure correct pin mode is set for ESP32, matching the behavior for other architectures.
* Modified servo control in `MFServo.cpp` and `MFServo.h` to use the `ESP32Servo` library and set appropriate pulse ranges for ESP32 boards.

**Build and Firmware Handling**

* Adjusted the firmware copying script (`copy_fw_files.py`) to only convert `.bin` files to `.uf2` format for Raspberry Pi Pico boards, preventing unnecessary conversion for ESP32.

These changes collectively enable the project to build and run on the ESP32-S3 platform, while maintaining compatibility with existing supported architectures.


miathedev:
Did i understand it correctly that i require an "mfmc" file to let MogiFlight correctly detect IOs?
Can someone please help me create one?

Product Link to the board:
https://heltec.org/project/wireless-stick-lite-v2/

elral:
> Did i understand it correctly that i require an "mfmc" file to let MogiFlight correctly detect IOs?  
> Can someone please help me create one?

First the easy part ;)  
No, an `mfmc` file is not required to detect a board by the connector. The defined `type` in the FW and the json file must match. It looks like you have this already considered. Once the board is recognized, you can define and upload a config (which device on which pin) and save additionally this configuration in a `mfmc` file.

elral:
> ESP32-S3 Board Support

Thanks for contributing your work to this repo!

But 'unfortenutely' ESP32 boards are not the boards we are planing to support officially. There are so many different variants of ESP32 boards which makes it not possible to support all of them. If we decide for one or two boards, I bet these are the wrong ones.  
Our fokus on boards with 32bit architecture are clearly the Pico boards from Raspberry. For now there is the Raspberry Pi Pico board officially supported, it might be that the Pico2 will also be supported in the near future.

So I don't think that this PR will get approved, but Sebastian has to decide it.

Nevertheless, let me comment your great PR.

*   copy\_fw\_files.py: I would check for `platform = env.BoardConfig().get("platform", {})` and not check for the name
*   I would run a script to create a combined .bin FW file which makes it easier to flash outside of VScode
*   A definition of `-DBUFFER_LENGTH=I2C_BUFFER_LENGTH` in the platformio.ini would be sufficient to solve the issue with the LCD library. That could be also done for other architectures which would need other definitions without changing the library.
*   Is there a reason not to use `ArduinoUniqueID` ? I had no problem using it with some ESP32 boards.
*   `pinMode(_pin, OUTPUT);` is the same as for all boards except for the Pico, no need to differ
*   The deviceBuffer should also be defined with `std::size_t`
*   Support fot the 2nd Core is missing
*   AnalogRead() value must be devided by 4
*   My experience is that the ESP32 shift out/in the data too fast and it has to slow down.
*   Flashing the boards via the connector has to be added.

You can also have a look on this branch [https://github.com/elral/MobiFlight-FirmwareSource/tree/ESP32_support](https://github.com/elral/MobiFlight-FirmwareSource/tree/ESP32_support) to find more informations of what I have listed.

DocMoebiuz:
Hi, thank you for your effort! Unfortunately, we decided we won't officially support ESP32. Like Ralf said, there are just too many and it would mean too much extra effort for the core team.

Please be encouraged to maintain a custom firmware. It can also be shared with the community, hopefully in the future we will have a registry for these things so that others can find it. If you need help to set up the custom firmware repository, come to our discord and we can try to help you there.

I will close this PR.

PR #376, opened 2026-03-31
Disable DTR on connect

elral:
## Description of changes

When connecting to AVR boards with DTR enabled, these boards get resettet on connect. After resetting the boards respond at least with `17,OK;` (not processed from the connector). Other board types which are used for community devices might also respond with additional messages which could lead to an unrecognized board or set them into bootlaoder mode (could happen for some ESP32 boards).

This PR disables setting DTR line on connect for all boards except for ProMicro and Pico which requires setting the DTR line.

See also PR #2870 and issue #2849 in the connector repo.

### DIFF
--- a/_Boards/arduino_mega.board.json
+++ b/_Boards/arduino_mega.board.json
@@ -11,7 +11,7 @@
     "ExtraConnectionRetry": false,
     "ForceResetOnFirmwareUpdate": false,
     "DelayAfterFirmwareUpdate": 0,
-    "DtrEnable": true,
+    "DtrEnable": false,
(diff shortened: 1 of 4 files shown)


ccrawford:
Thank you! Community device authors will be pleased.

elral:
I will close this PR as the plain Arduino Mega with 32u4 chip on board as USB-seriell converter needs the DTR line enabled to work.
If we want to disable the DTR line for CH340 chips, we have to set up two different board.json files and figure out all VID/PID's for CH340 chip.

PR #377, opened 2026-06-24
include png files in zip file

elral:
## Description of changes

png files are not copied to the zip file, this is fixed with this PR.


DocMoebiuz (approved):
Thanks!

PR #378, opened 2026-06-24
Fix pico2 support

elral:
## Description of changes

- [x] a new reset file is provided which jumps into the bootloader mode after clearing the EEPROM.
- [x] both pico png images have the correct file name


DocMoebiuz (approved):
Nice!

